PR #1, opened 2020-09-19
Add insertion op; improve deletion op; add new way to simplify constants

MilesCranmer:
Insertion op:
- Insert random operator inside an equation at a random node, rather than just at a leaf. Binary trees will add a new constant/variable on one of their branches.

Deletion op:
- Now when deleting an operator from the middle of an equation, it will leave the enclosed operations, rather than replacing them with a constant/variable.

Simplification op:
- ((x0 + const) + const)=>(x0+const), and ((x0 * const) * const) => (x0 * const) after initially simplifying all constant parts of an equation, like (const ... const) => (const). This seems to help a lot with the new insertion op, which seemed to result in many constants being added together.

Hyperparams:
- These have been set to hand-chosen values until a new hyperparameter optimization is run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,11 @@ which gives:
 ### API
 
 What follows is the API reference for running the numpy interface.
-Note that most parameters here
-have been tuned with ~1000 trials over several example
-equations, so you likely don't need to tune them yourself.
+You likely don't need to tune the hyperparameters yourself,
+but if you would like, you can use `hyperopt.py` as an example.
 However, you should adjust `threads`, `niterations`,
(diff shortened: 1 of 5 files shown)


MilesCranmer:
E.g., 

```
julia> printTree(n)
plus(plus(x0, 1.0), 2.0)
julia> printTree(insertRandomOp(copyNode(n)))
div(plus(plus(x0, 1.0), 2.0), 1.2341375)
julia> printTree(combineOperators(copyNode(n)))
plus(x0, 3.0)
```

PR #4, opened 2020-09-27
Add capabilities of running on many-node cluster via slurm

MilesCranmer:
Example given with `slurm_example.py`. This will need to be modified to the cluster, as it calls `sbatch ...` itself.

I will need to do some final checks before merging. Right now, the speed is definitely very high - I can get 10^7 cycles per second. But I should check that the hall of fame equation array is actually being shared between nodes.

- [ ] Improve docs on how to run/modify.
- [ ] Potentially split up files into a more sensible directory structure, and so `slurm_example.py` doesn't actually contain slurm code.
- [ ] Assert that hall of fame and sub-populations are being shared and migrated between nodes.
- [ ] Test that it can be run for several hours without crashing on many nodes.
- [ ] Debug why it takes so long to start the program. Is this normal for Julia? Or is something wrong here?


MilesCranmer:
Closing for now; multi-node stuff will be easier once I switch PySR to use the `SymbolicRegression.jl` version.

However, in general it will be much easier to configure this in the pure-Julia version, e.g., it can be done with:
```julia
using ClusterManagers

addprocs(SlurmManager(2))
addprocs()
```

instead of just `addprocs()` as given in the README example.

I think for the Python side, I'll just wrap a couple basic options for Slurm, but more complex clusters should just configure the Julia one.

PR #24, opened 2021-01-17
Refactored sr.jl and sr.py heavily

DhananjayAshok:
Split sr.jl into many composite files and used the include tag for increased readability and "modularity"

- Similar functions have gone to the same file
- Structs and functions defined specially on those structs have gone to the same file
- long functions have been given a function of their own

Performance Benchmark on main.py:
- Before refactoring main.py terminates between 38-42 seconds
- After refactoring main.py terminates between 37 - 41 seconds (no significant difference)


julia/randomMutations.jl
```diff
@@ -227,4 +227,13 @@ function deleteRandomOp(tree::Node)::Node
         end
     end
     return tree
+end
+
+# Create a random equation by appending random operators
+function genRandomTree(length::Integer)::Node
```
DhananjayAshok, line 233: genRandomTree can work using the functions defined up to this point, and is a natural extension of all other random tree functions in this file

julia/Node.jl
```diff
@@ -139,4 +139,35 @@ function countConstants(tree::Node)::Integer
     else
         return 0 + countConstants(tree.l) + countConstants(tree.r)
     end
+end
+
+# Get all the constants from a tree
```
DhananjayAshok, line 144: These two tree constant functions can work if defined earlier as well, they also do not rely on anything other than a tree and so can be considered to be a Node function just like the others

julia/utils.jl
```diff
@@ -8,4 +8,25 @@ end
 
 function getTime()::Integer
     return round(Integer, 1e3*(time()-1.6e9))
+end
+
+# Check for errors before they happen
+function testConfiguration()
```
DhananjayAshok, line 14: This only requires constants that are defined in constants.jl to work and is a utilities function

julia/sr.jl
```diff
@@ -27,6 +27,10 @@ include("simplification.jl")
 
 include("PopMember.jl")
 
+
+include("halloffame.jl")
```
DhananjayAshok, line 31: Moved halloffame definition up to right after PopMember as this is the soonest it can be defined

pysr/sr.py
```diff
@@ -447,7 +447,7 @@ def pysr(X=None, y=None, weights=None,
     global_extra_sympy_mappings = extra_sympy_mappings
 
     print("Running on", ' '.join(command))
-    process = subprocess.Popen(command, stdout=subprocess.PIPE, bufsize=1)
+    process = subprocess.Popen(command, stdout=subprocess.PIPE, bufsize=1, shell=True)
```
DhananjayAshok, line 450: on a Windows System for some reason without the shell argument I get: 
FileNotFoundError: [WinError 2] The system cannot find the file specified

Solution from: https://stackoverflow.com/questions/42572582/winerror-2-the-system-cannot-find-the-file-specified-python

MilesCranmer:
Very nice @DhananjayAshok!! Thank you for the fantastic refactoring job! I really like how clean this looks, nice work. I will review and test and let you know if I have any questions.

By the way, I have also created a pure Julia version here: https://github.com/MilesCranmer/SymbolicRegression.jl.

I think I can include lot of your changes to the Julia code in that package as well, since it stems from the same git history and is nearly an identical Julia codebase. Let me merge your changes to that one offline (lots of changes to the function signatures - I think I can automate these); it shouldn't be too much work.

Cheers, and thanks again Dhananjay! 
Miles

MilesCranmer:
Pushed some changes, let me know what you think!

1. Changed the remaining filenames to be in CamelCase with full words.
2. Moved the `include()` from the Julia code to the Python metaprogramming part. This is required to set up the Julia call from Python, since Julia runs the code in `/tmp` and needs to know where the code is. (I also do it with the `operators.jl` file).
3. Moved the imports to the files that use them

MilesCranmer:
Actually, do you want to submit a second PR for `lgga/`? Then I can merge the PySR refactoring first, and we can discuss the `lgga/` stuff after?

Cheers and thanks again!
Miles

DhananjayAshok:
Hey Miles, 
Thanks for your review and comments!

I have fully removed the LGGA functionality from this branch and will add it back in to a new lgga branch.
I have also refactored sr.py a bit. Essentially I split up the pysr function into many smaller function calls in the hope that it makes the code easier to work with down the line. 

Have tested this code at this version and it still seems to be performing the exact same on my system. 

Do let me know if you have any questions or comments.
Thanks,
Dhananjay

pysr/sr.py
MilesCranmer: Note that I have continuous testing on windows which works fine without this argument: https://travis-ci.com/github/MilesCranmer/PySR/jobs/466146940, so I think the FileNotFoundError on your system could be from something else.

This change seems to be causing my PySR to hang. I think this `shell=True` argument is what is breaking it. `shell=True` means you are passing a single string, but `shell=False` means you are passing a list - which PySR does. So we will need to change this back.

DhananjayAshok:
I think there is a bit of an issue here:
>> python example.py
.....
.....
```
Traceback (most recent call last):
  File "C:\Users\dhana\Desktop\SelfLearn\LGGA\PySR\example.py", line 9, in <module>
    equations = pysr(X, y, niterations=5,
  File "C:\Users\dhana\Desktop\SelfLearn\LGGA\PySR\pysr\sr.py", line 269, in pysr
    _set_globals(**kwargs)
TypeError: _set_globals() missing 1 required positional argument: 'extra_sympy_mappings'
```

MilesCranmer:
Sorry, one second. I am almost done with my edits.

DhananjayAshok:
No, my bad I think I introduced a bug by redefining extra_sympy_mappings twice. I'll hold off on any edits till you finish your work and merge the branch into master to avoid any other issues.

DhananjayAshok:
If possible could you drop me an email when you are done with your changes. I'm planning on holding off on creating the LGGA branch until the merge to avoid conflicts.

Thanks!

MilesCranmer:
No worries at all, I should have mentioned I was making edits. And that sounds good!

Cheers,
Miles

PR #25, opened 2021-01-19
Added Feynman Equations Dataset and Problems for Testing

DhananjayAshok:
The 100 Feynman Equations from Physics is a set of equations of varying complexity that is a good benchmark for a Symbolic Regression tool. AI Feynman in their paper uses this as a benchmark as well. I have added a Problem file with a generic problem class and made Feynman Problems a child of this problem class. 

I have made it such that you can very quickly test pysr on a FeynmanEquation of your choosing.

This can be done from the root directory in a python shell with:
```
>>> from pysr.sr import pysr, best
>>> from pysr.Problems import FeynmanProblem
>>> problem_list = FeynmanProblem.mk_problems(first=5, gen=True, dp=200)
<string>:1: RuntimeWarning: invalid value encountered in sqrt
>>> problem = problem_list[0]
>>> print(problem)
Feynman Equation: I.6.2a|Form: exp(-theta**2/2)/sqrt(2*pi)
>>> equations = pysr(problem.X, problem.y, variable_names=problem.var_names)
...
Cycles per second: 1.020e+05
Progress: 399 / 400 total iterations (99.750%)
Hall of Fame:
-----------------------------------------
Complexity  MSE        Score     Equation
0           5.972e-03  0.000e+00  0.020923
8           1.216e-13  3.077e+00  mult(exp(mult(mult(theta, theta), -0.5000038)), 0.3989443)
10          1.340e-14  1.103e+00  mult(exp(mult(plus(mult(theta, theta), -0.06846976), -0.50000024)), 0.38551515)

>>> print(best(equations), " vs ", problem.form)
0.398941738836597*exp(-0.50000024*theta**2)  vs  exp(-theta**2/2)/sqrt(2*pi)
```

Alternately one could also run 

```
>>> from pysr.Problems import FeynmanProblem, run_on_problem
>>> problem_list = FeynmanProblem.mk_problems(first=5, gen=True, dp=200)
<string>:1: RuntimeWarning: invalid value encountered in sqrt
>>> problem = problem_list[0]
>>> print(problem)
Feynman Equation: I.6.2a|Form: exp(-theta**2/2)/sqrt(2*pi)
>>>best_equation, real_equation, others = run_from_problem(problem)

```


MilesCranmer:
Sorry for the delay in reviewing this; will try to get to it next week! But in general, this looks like an amazing addition!!! Thanks for another great PR :)
Cheers,
Miles

MilesCranmer:
Finally merged this. Thanks again for the awesome work!!

Here are some small changes I made to the merged copy:

- `pysr.Problems.FeynmanProblem` -> `pysr.FeynmanProblem`
- OS-compatible filenames
- Have numpy automatically convert to 32-bit float to avoid complaints from Julia when parsing
- Fixed some bugs on my side in passing variables with names to the Julia code

Thanks @DhananjayAshok!!
Cheers,
Miles

dbl001:
In FeynmanProblem.py on line #110 in the function: run_on_problem(), please replace:

```
return str(best(equations)), problem.form, others
```
with
```
  return str(self.get_best(equations)), problem.form, others
```

dbl001:
This code change in 'feynman_problems.py' resolves the NaN problem while generating datasets.
Instead of populating the dataset from a uniform distribution between 0.01 and 25, it uses a uniform distribution 
between the high and low values specified in:

https://github.com/MilesCranmer/PySR/blob/master/datasets/FeynmanEquations.csv#L92

before:
```
            self.X = np.random.uniform(0.01, 25, size=(self.dp, self.n_vars))
```
after:
```
            self.X = np.empty((self.dp, self.n_vars))
            for i in range(self.n_vars):
                self.X[:, i] = np.random.uniform(self.low[i], self.high[i], size=self.dp)

```

MilesCranmer:
Thanks @dbl001! Would you please submit this as a PR so you can be credited for the fix?

dbl001:
Ok.  Currently running through all the Feynman test with: do_feynman_experiments() to verify correctness.

PR #26, opened 2021-01-20
Put Julia code into separate package

MilesCranmer:
This is a PR to call the package `SymbolicRegression.jl` instead of storing the julia code locally in this repo.

- [x] Test inclusion of weights
- [x] Test that we can use different number of processes
- [x] Test if performance changes
- [x] Add mechanism to include local SymbolicRegression.jl folder instead of package import, so can easily develop the Julia backend
- [x] Assert that inline functions still work
- [x] Debug appearance of `nothing` in symbolic simplification. Could be due to limited selection of operators? In which case I can just turn it off...

FYI @DhananjayAshok


MilesCranmer:
@DhananjayAshok - I enabled local SymbolicRegression.jl installs, so it's possible to work on the Julia code, and test it with the Python frontend!

For example, to use the SymbolicRegression.jl repo in `../SymbolicRegression.jl`, use:
```python
equations = pysr(X, y, niterations=5,
    binary_operators=["plus", "mult", "sub", "div"],
    unary_operators=["cos", "exp"],
    local_install="../SymbolicRegression.jl")
```
If installed with `import Pkg; Pkg.install(...)`, it will use the installed version unless this `local_install` flag is set.

It assumes the git repo is set to this PR: https://github.com/MilesCranmer/SymbolicRegression.jl/pull/6.

Issue #35, opened 2021-03-04
Allow `y` to be vector-valued?


patrick-kidger:
It would be nice to allow `y` to be vector valued, essentially as a (parallelised?) shortcut for something like the following:

```python
expressions = []
for index in range(y.shape[-1]):
    yi = y[:, index]
    expression_i = pysr.pysr(x, yi)
    expressions.append(expression_i)
```

This links to (and is actually motivated by) my suggestion in #32: `as_pytorch` could then return a single `SymPyModule` wrapping a list of all `expressions`, rather than returning several `SymPyModules` for each one, which would then have to be composed together in the above for loop.

MilesCranmer:
Definitely agree. Actually it could be even nicer to just have the SymbolicRegression.jl backend loop over each y-component and return a list of equations at each complexity. Will think more about this.

MilesCranmer:
Just added capabilities for multi-output (and multiple equations returned). SymbolicRegression.jl will now return a list of equations for each output. 

This should be simple enough to set up for just importing into PySR. The trickier part is the JAX and PyTorch backend! I'm not sure how to approach - whether to continue outputting a different Torch/JAX model for each equation, or try to merge? It's probably overcomplicated to merge...

patrick-kidger:
Is the PyTorch backend (you mean exporting via `sympytorch`?) actually available? I don't see it based on a quick glance through the code.

For sympytorch at least it's possible to pass multiple equations to the `SymPyModule`, i.e. vector values are automatically supported.

MilesCranmer:
The PyTorch backend hasn't been implemented yet; it's next on my todo list!

That is nice that vector equations are supported! It doesn't assume the equations are all the same, but for different parameter choices, right? They can be completely different equations?

patrick-kidger:
Yep, completely different equations.

MilesCranmer:
Actually on second though I'm not sure it will work due to the API of PySR: it will output a list of equations for each feature of `y`. When JAX/Torch export is turned on, it will generate a separate function/module for each expression. But I think the user will want to filter equations for each feature of `y` separately before making a choice of equation, and then fusing them.
I don't see an easy way to do this otherwise? I also don't know if there'd be a speedup since the expressions are assumed to be different, and no operations are re-used?

patrick-kidger:
[...]

PR #48, opened 2021-05-31
PyTorch export

MilesCranmer:
This uses (a slightly-modified version of) @patrick-kidger's [sympytorch](https://github.com/patrick-kidger/sympytorch) to export discovered equations to PyTorch. Parameters in the module are set to the default as found by PySR, but are trainable.

This essentially allows one to do the deep learning -> symbolic regression technique from https://github.com/MilesCranmer/symbolic_deep_learning, but plug the discovered equation back into the network, without any manual work.

This is the same format as the JAX export - it expects a matrix `X` as input with columns corresponding to symbols, except it outputs a PyTorch module with embedded parameters instead of a (function, parameters) tuple.

@patrick-kidger, what do you think of this way of doing things? The modifications from your package include: expects matrix X as input rather than different vector for each symbol (to unify with the JAX backend) and only one expression per module (was unclear how to merge multiple expressions which the user may or may not use).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,16 +61,22 @@ jobs:
             python setup.py install
       - name: "Install Coverage tool"
         run: pip install coverage coveralls
+      - name: "Run tests"
+        run: coverage run --source=pysr --omit='*/feynman_problems.py' -m unittest test.test
+        shell: bash
       - name: "Install JAX"
         if: matrix.os != 'windows-latest'
         run: pip install jax jaxlib # (optional import)
         shell: bash
-      - name: "Run tests"
-        run: coverage run --source=pysr --omit='*/feynman_problems.py' -m unittest test.test
-        shell: bash
       - name: "Run JAX tests"
         if: matrix.os != 'windows-latest'
         run: coverage run --append --source=pysr --omit='*/feynman_problems.py' -m unittest test.test_jax
+        shell: bash
+      - name: "Install Torch"
+        run: pip install torch # (optional import)
+        shell: bash
+      - name: "Run Torch tests"
+        run: coverage run --append --source=pysr --omit='*/feynman_problems.py' -m unittest test.test_torch
         shell: bash
       - name: Coveralls
         env:
--- a/pysr/__init__.py
+++ b/pysr/__init__.py
@@ -1,3 +1,3 @@
 from .sr import pysr, get_hof, best, best_tex, best_callable, best_row
 from .feynman_problems import Problem, FeynmanProblem
-from .export import sympy2jax
+from .export_jax import sympy2jax
--- a/pysr/__init__.py
+++ b/pysr/__init__.py
@@ -1,4 +1,2 @@
 from .sr import pysr, get_hof, best, best_tex, best_callable, best_row
 from .feynman_problems import Problem, FeynmanProblem
-from .export_jax import sympy2jax
-from .export_torch import sympy2torch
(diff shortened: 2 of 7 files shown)


patrick-kidger:
Seems reasonable as is. Although it may be worth leaving this as a separate package as per my comments in #35? No strong feelings.

MilesCranmer:
Thanks!

Yeah, longterm it will probably be best to have a separate modular package for generic SymPy -> trainable modules for numpy/torch/jax/etc. But I guess in the short-term, PySR exports are a bit implementation-specific (for consistency, want an array `X` as input, rather than inputting vectors for each symbol like `sympy.lambdify` does), and as a PySR user, I'd rather just have it do the exporting itself without an extra package. Thoughts?

patrick-kidger:
No strong feelings tbh.
I'd note that if you want an array then the approach I'd take would be something like
```python
class PySRSymPyModule(torch.nn.Module):
    def __init__(self, **kwargs):
        super().__init__()
        self._module = sympytorch.SymPyModule(**kwargs)

    def forward(self, X):
        symbols = {f"x{i}": xi for i, xi in enumerate(X)}
        return self._module(**symbols)
```
to simply wrap existing functionality rather than duplicating it wholesale.

MilesCranmer:
You are right; this is probably a cleaner way to do it. I tried implementing this on `sympytorch-export`.

However, I realized an issue: there is no good way to deal with the install of the exporter libraries. There's two tricky scenarios:
- User doesn't want torch(/jax) export. So sympy->torch(/jax) and torch(/jax) shouldn't be installed.
- User does want torch(/jax) export. But they only install torch(/jax) and PySR, not realizing they need a third library for exporting.

I think unfortunately the only good way to deal with these is to put the sympy->torch code here (as in this pull request) along with the sympy->jax code, then do a lazy initialization only if the exporter code is called...

patrick-kidger:
So I think that should be doable as well. Pseudocode:
```python
def export(obj_to_export):
    try:
        import export_library
    except ImportError as e:
        raise ImportError("Please additionally install `export_library` to export to `deep_learning_library`.") from e
    return export_library.export(obj_to_export)
```
If you want you could additionally declare `export_library` as an [optional dependency](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies).

MilesCranmer:
That sounds like a great way to do it. So, can one get `setup.py` to install `sympytorch` if `torch` is already installed? (But not otherwise - which would trigger a torch install)?

MilesCranmer:
Maybe this?

```python
import setuptools
import sys

...

setuptools.setup(
    ....
    install_requires=( [
            "numpy",
            "pandas",
            "sympy"
            ] + (
            ['git+https://github.com/patrick-kidger/sympytorch']
            if 'torch' in sys.modules else []
            )
        ),
    ...
)
```

MilesCranmer:
By the way, are you planning on putting sympytorch on PyPI?

MilesCranmer:
Argh, found some more problems that are specific to PySR. I'm just going to merge the implementation-specific version and try to set up an integration with a generic package later. Maybe for a jax and numpy exporter too. (Maybe all in one repo! Since sympy.lambdify doesn't let one get the parameters out)

Also, can the 3.8 requirement for Python be reduced in sympytorch or is it needed?

patrick-kidger:
Worth noting that this:
```
['git+https://github.com/patrick-kidger/sympytorch'] if 'torch' in sys.modules else []
```
won't work as that only checks if `torch` has been imported, not installed.

At your request, I've just put sympytorch on PyPI, and loosened the requirements to Python 3.7. Let me know if it all seems to work for you-- if it does I'll update the install instructions.

MilesCranmer:
Thanks!! This seems to have fixed all the issues! I now have sympytorch working on the master branch for PyTorch exports 👍

PR #87, opened 2022-01-14
Switch Julia interface to PyJulia

MilesCranmer:
This switches the entirety of the Julia/Python interface to use PyJulia rather than a `subprocess` call to a generated script. Multithreading and distributed computing both work! 

This brings several advantages:

- Huge speedups in startup time on second `pysr( )` calls in the same python call
- Simplified error debugging on the Julia side
- Greatly simplified codebase with no huge chunks of code generation
- Possibility to checkpoint the symbolic regression search, and restart later

The disadvantage is that this will put PySR at the mercy of PyJulia and any bugs that it experiences. However, I think in the longterm this is still much better than running a script of generating code.

Installation with this new interface is now:
```bash
pip install pysr
python -c 'import pysr; pysr.install()'
```
This will install PyCall.jl (required by PyJulia) as well as the packages defined in `SymbolicRegression.jl`.

Will initially be implemented in v0.7.0a1 for a while until things stabilize.

TODO:

- [x] Update documentation to describe PyJulia install.
- [x] Update documentation to mention that repeat calls will have a quicker startup time.
- [x] Check that error is correct when PyJulia not installed if `pysr.install()` has not been run.
- [x] Check if there are any performance hits using PyJulia vs Julia.
- [x] Make it so we can KeyboardInterrupt the PyJulia execution of EquationSearch. See https://github.com/JuliaPy/pyjulia/issues/211.
- [x] This does not work when python is statically linked. In this situation, should we use `compiled_modules=False`? Or enfoce that the user has to get this working?
- [x] Is there anything I can do to get conda working with PyCall.jl?
- [x] Test conda with the CI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,6 +61,7 @@ jobs:
             python -m pip install --upgrade pip
             pip install -r requirements.txt
             python setup.py install
+            python -c 'import julia; julia.install()'
       - name: "Install Coverage tool"
         run: pip install coverage coveralls
       - name: "Run tests"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,7 +61,7 @@ jobs:
             python -m pip install --upgrade pip
             pip install -r requirements.txt
             python setup.py install
-            python -c 'import julia; julia.install()'
+            python -c 'import pysr; pysr.install()'
       - name: "Install Coverage tool"
         run: pip install coverage coveralls
       - name: "Run tests"
--- a/.github/workflows/CI_Windows.yml
+++ b/.github/workflows/CI_Windows.yml
@@ -61,7 +61,7 @@ jobs:
             python -m pip install --upgrade pip
             pip install -r requirements.txt
             python setup.py install
-            python -c 'import julia; julia.install()'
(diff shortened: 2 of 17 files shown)


MilesCranmer:
To get KeyboardInterrupt to work, I'm going to launch EquationSearch in a `multiprocessing.Process`. Upon KeyboardInterrupt, a signal will be given to a temp file. The actual SymbolicRegression.jl EquationSearch will check for this signal during the search loop. It will then manually exit if this signal is set to 1.

MilesCranmer:
So it seems like PySR actually speeds up with PyJulia versus the old method of launching a script. 

Measurements before the change:
```
[64700., 67500., 68800., 68900., 69300., 69800., 68400., 73000.,
       74600., 75100., 75500., 75400., 57700., 59700., 60600., 61300.,
       61900., 62500., 62900.]
```
(average 67242)
Measurements after the change:
```
[ 87800.,  87900.,  88000.,  87700.,  87200.,  90900.,  90800.,
        89500.,  89000.,  88700., 121000., 121000., 121000.]
```
(average 96192).

These measurements are very small jobs but this at least shows it doesn't drastically hurt performance, so I think we are good on that front.

Issue #27, opened 2021-02-03
Julia registry updating slowly


MilesCranmer:
FYI I pushed v0.4.2 of SymbolicRegression.jl to the registry 15 hours ago, which is required for the latest PySR. However, the registry server is still not updating - which seems like an issue that sometimes happens: https://github.com/JuliaRegistries/General/issues/16777.

To get the Julia registry to stay up to date even if the registry server fails, you can use the git version instead. This can be done as follows:

1. Delete your registry folder `~/.julia/registries/General`
2. Launch Julia.
3. Run the following commands:
```julia
ENV["JULIA_PKG_SERVER"] = ""
import Pkg
Pkg.update()
```

This will install the git version of the registry, which is always up-to-date.

MilesCranmer:
Server seems to have updated. Closing.

MilesCranmer:
Unfortunately appearing again for v0.5.10. I have pinned this issue in case some of the Julia servers are out of date - just follow the above commands to get it working.

MilesCranmer:
Thanks!

I wonder if I should have PySR specify the `Manifest.toml` file (which could indicate the git repo) rather than just the `Project.toml` (which only indicates the project hash)? Then I suppose this would never be an issue - could just point directly to SymbolicRegression.jl.

mkitti:
If this is meant as a library, I would not include the Manifest.toml.

I typically only include the Manifest.toml for terminal projects that I may want to reproduce exactly.

MilesCranmer:
Oh I meant for the point about including SymbolicRegression.jl based on the git repo URL, rather than the Julia registry. I'm not sure how else to do that, for PySR users, other than to include a fixed Manifest.toml. (is there any way to enforce that in a Project.toml?)

mkitti:
How users install your package is up to them not you,  ultimately. For some, they either have to use specific mirrors or proxies for reliable access in their geographic locations. This is the problem that the pkg server protocol solves by providing a way to mirror and verify packages in a distributed and versioned fashion.

Pointing to your specific repository is a secondary option for those who cannot wait for the pkg servers to catch up and who also have reasonable direct access to Github.

So for the kind of package I think you are providing, I would recommend not including the Manifest.toml.

MilesCranmer:
> How users install your package is up to them not you, ultimately

[...]

PR #107, opened 2022-02-13
Install SymbolicRegression directly from GitHub repo

MilesCranmer:
This would finally eliminate #27, as PySR would install the backend directly from GitHub, rather than through the Julia registry. No need to wait for the registry to update.

@mkitti does this look like what you had in mind?

(note that I use the temp dir for other files, so create it manually, rather than using the `temp` argument to Pkg.)


pysr/sr.py
```diff
@@ -337,25 +334,13 @@ def init_julia():
...
-
-    project_toml_path = tmp_dir / "Project.toml"
-    project_toml_path.write_text(project_toml)
+def _add_sr_to_julia_project(Main, io_arg):
+    Main.spec = Main.PackageSpec(
+        name="SymbolicRegression",
+        url="https://github.com/MilesCranmer/SymbolicRegression.jl",
+        rev="v0.7.8",
+    )
+    Main.eval(f"Pkg.add(spec, {io_arg})")
```
mkitti, line 343: This confuses me. What is this second `io_arg` argument?
https://pkgdocs.julialang.org/v1/api/#Pkg.add

mkitti: Ah, I see https://pkgdocs.julialang.org/v1/api/#Redirecting-output

MilesCranmer: Just to add: the reason I have to do the `Main.eval` here is because the PyObject wrapper for IOStreams doesn't seem to pass correctly from Python. It's a bit ugly but works for now. The reason I pass the full `io_arg` instead of something like `io={io}` is because the IO redirection was only implemented in Julia 1.5+.

MilesCranmer:
I did not know you could do that, thanks! I will switch to using that instead.

Any idea how to pass this to `Pkg.activate` as a variable? I see two different outputs currently.

with `]`
```julia
(@v1.8) pkg> activate @pysr
  Activating new project at `~/.julia/environments/pysr`
```

and manual call:
```julia
julia> Pkg.activate("@pysr")
  Activating new project at `~/Documents/PySR/@pysr`
```

MilesCranmer:
This seems like it shouldn't be the expected behavior so opened an issue on Pkg.jl: https://github.com/JuliaLang/Pkg.jl/issues/2983

MilesCranmer:
Okay, the environment is now given by "pysr-{__version__}", rather than created as a temp directory. This is much cleaner in my opinion. Thanks for the idea @mkitti!

mkitti:
With the conda-forge version, `.julia` should be a folder located within the `$CONDA_PREFIX`.

MilesCranmer:
Great!

I think the `shared = true` will automatically find that depot in conda. Let me check.

mkitti:
Yes, it should. We set it up for that purpose.

PR #117, opened 2022-03-10
Update PySR default settings

MilesCranmer:
This will be for PySR v0.8.0. These settings have been tuned over ~80,000 trials on two hard test problems that PySR had a hard time with originally, as discussed here: https://github.com/MilesCranmer/PySR/discussions/115.

This also includes the adaptive parsimony update in the backend.


MilesCranmer:
I may want to update the default `niterations`, as the new settings are potentially too short.

PR #133, opened 2022-05-08
Enable custom cluster managers

MilesCranmer:
This let's PySR launch jobs over a cluster. If you are on a slurm cluster (on the head node of a job), you simply pass `cluster_manager="slurm"` to PySRRegressor, and it will set up a bunch of workers over the slurm job. Likewise, the other cluster managers listed on https://github.com/JuliaParallel/ClusterManagers.jl can be used.

The one unfortunate thing is I don't think this functionality can be unit-tested; it will just be manual tests on clusters.

cc @KazeWong!

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -333,12 +333,18 @@ def init_julia():
 
 
 def _add_sr_to_julia_project(Main, io_arg):
-    Main.spec = Main.PackageSpec(
+    Main.sr_spec = Main.PackageSpec(
         name="SymbolicRegression",
         url="https://github.com/MilesCranmer/SymbolicRegression.jl",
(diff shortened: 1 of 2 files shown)


MilesCranmer:
This model:
```python
model = PySRRegressor(
     unary_operators=["cos"],
     procs=40 * 4,
     populations=40 * 4 * 2,
     population_size=1000,
     ncyclesperiteration=3000,
     cluster_manager="slurm",  # < Only change required
)
```
will launch 160 workers over a slurm process (4 nodes in this case).

Here's what the usage looks like:

Screen Shot 2022-05-07 at 10 30 40 PM

Keep in mind this was launched from within IPython on the head node!

PR #156, opened 2022-07-02
[Feature] LaTeX table generator

MilesCranmer:
This generates a booktabs-style LaTeX table for a subset of equations. Here is an example:

```python
import numpy as np

X = 2 * np.random.randn(100, 5)
y = 2.5382 * np.cos(X[:, 3]) + X[:, 0] ** 2 - 0.5

model = PySRRegressor(
    niterations=80,
    binary_operators=["+", "*"],
    unary_operators=["cos"],
    model_selection="best",
    loss="loss(x, y) = (x - y)^2",  # Custom loss function (julia syntax)
    maxsize=11,
)

model.fit(X, y)

print(model.latex_table(precision=3, include_score=True))
```

The output of this is:
```latex
\begin{table}[h]
\begin{center}
\begin{tabular}{@{}lccc@{}}
\toprule
Equation & Complexity & Loss & Score \\
\midrule
$3.9$ & 1 & 38.9 & 0 \\
$x_{0}^{2}$ & 3 & 3.16 & 1.26 \\
$x_{0}^{2} - 0.257$ & 5 & 3.09 & 0.0105 \\
$x_{0}^{2} + \cos{\left(x_{3} \right)}$ & 6 & 1.26 & 0.898 \\
$x_{0}^{2} + 2.44 \cos{\left(x_{3} \right)}$ & 8 & 0.245 & 0.818 \\
$x_{0}^{2} + 2.54 \cos{\left(x_{3} \right)} - 0.5$ & 10 & 2.28e-13 & 13.9 \\
\bottomrule
\end{tabular}
\end{center}
\end{table}
```
which renders as:
image

Leaving `include_score` set to `False` will leave out the Score column. Precision can be adjusted to have more or less precise constants.

One can render only a subset of equations by using `latex_table([1, 4])` which only includes the 1st and 4th equation in `model.equations_`.

---

Edit: it now renders the `e-13` as `\cdot 10^{-13}`

### DIFF
--- a/pysr/export_latex.py
+++ b/pysr/export_latex.py
@@ -6,21 +6,24 @@ from sympy.printing.latex import LatexPrinter
 class PreciseLatexPrinter(LatexPrinter):
     """Modified SymPy printer with custom float precision."""
 
-    def __init__(self, settings=None, prec=3):
+    def __init__(self, settings=None, prec=3, full_prec=True):
         super().__init__(settings)
         self.prec = prec
+        self.full_prec = full_prec
 
     def _print_Float(self, expr):
         # Reduce precision of float:
-        reduced_float = sympy.Float(expr, self.prec)
+        reduced_float = sympy.Float(expr, self.prec, full_prec=self.full_prec)
         return super()._print_Float(reduced_float)
 
 
-def to_latex(expr, prec=3, **settings):
+def to_latex(expr, prec=3, full_prec=True, **settings):
     """Convert sympy expression to LaTeX with custom precision."""
     if len(settings) == 0:
         settings = None
-    printer = PreciseLatexPrinter(settings=settings, prec=prec)
+    printer = PreciseLatexPrinter(
+        settings=settings, prec=prec, full_prec=full_prec
+    )
     return printer.doprint(expr)
 
 
--- a/pysr/export_latex.py
+++ b/pysr/export_latex.py
@@ -21,9 +21,7 @@ def to_latex(expr, prec=3, full_prec=True, **settings):
     """Convert sympy expression to LaTeX with custom precision."""
     if len(settings) == 0:
         settings = None
-    printer = PreciseLatexPrinter(
-        settings=settings, prec=prec, full_prec=full_prec
-    )
+    printer = PreciseLatexPrinter(settings=settings, prec=prec, full_prec=full_prec)
     return printer.doprint(expr)
 
 
--- a/pysr/export_latex.py
+++ b/pysr/export_latex.py
@@ -6,22 +6,20 @@ from sympy.printing.latex import LatexPrinter
 class PreciseLatexPrinter(LatexPrinter):
     """Modified SymPy printer with custom float precision."""
 
-    def __init__(self, settings=None, prec=3, full_prec=True):
+    def __init__(self, settings=None, prec=3):
         super().__init__(settings)
         self.prec = prec
-        self.full_prec = full_prec
 
     def _print_Float(self, expr):
         # Reduce precision of float:
-        reduced_float = sympy.Float(expr, self.prec, full_prec=self.full_prec)
+        reduced_float = sympy.Float(expr, self.prec)
         return super()._print_Float(reduced_float)
 
 
 def to_latex(expr, prec=3, full_prec=True, **settings):
     """Convert sympy expression to LaTeX with custom precision."""
-    if len(settings) == 0:
-        settings = None
-    printer = PreciseLatexPrinter(settings=settings, prec=prec, full_prec=full_prec)
+    settings["full_prec"] = full_prec
+    printer = PreciseLatexPrinter(settings=settings, prec=prec)
     return printer.doprint(expr)
 
 
--- a/pysr/export_latex.py
+++ b/pysr/export_latex.py
@@ -1,0 +1,34 @@
+"""Functions to help export PySR equations to LaTeX."""
+import re
+
+
+def set_precision_of_constants_in_string(s, precision=3):
+    """Set precision of constants in string."""
+    constants = re.findall(r"\b[-+]?\d*\.\d+|\b[-+]?\d+\.?\d*", s)
+    for c in constants:
+        reduced_c = "{:.{precision}g}".format(float(c), precision=precision)
+        s = s.replace(c, reduced_c)
+    return s
+
+
+def generate_top_of_latex_table():
+    latex_table_pieces = [
+        r"\begin{table}[h]",
+        r"\centering",
+        r"\caption{}",
+        r"\label{}",
+        r"\begin{tabular}{@{}lcc@{}}",
+        r"\toprule",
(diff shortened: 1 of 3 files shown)


MilesCranmer:
@tttc3 or @kazewong what do you think of this, and the API? Would any other options be useful here?

pysr/export_latex.py
```diff
@@ -0,0 +1,34 @@
+"""Functions to help export PySR equations to LaTeX."""
+import re
+
+
+def set_precision_of_constants_in_string(s, precision=3):
+    """Set precision of constants in string."""
+    constants = re.findall(r"\b[-+]?\d*\.\d+|\b[-+]?\d+\.?\d*", s)
+    for c in constants:
+        reduced_c = "{:.{precision}g}".format(float(c), precision=precision)
+        s = s.replace(c, reduced_c)
```
MilesCranmer, line 10: I'm realizing this might produce some bugs, if a constant is just something like `"1"` - then all the `1`s would be replaced...

Is there a better usage of `re` here?

MilesCranmer, line 10: I'm realizing this might produce some bugs, if a constant is just something like `"1"` - then all the `1`s would be replaced...

Is there a better usage of `re` here?

MilesCranmer: Should just write a custom `sympy.LatexPrinter` instead...

kazewong:
@MilesCranmer Do you want me to review this PR? If you assign it to me then I can have a look at this.

MilesCranmer:
Thanks! Added

MilesCranmer:
The weird inconsistency in the number of significant figures [here](https://github.com/MilesCranmer/PySR/pull/156/files#diff-ae070fbd66c55f7bf108994c0a42ffe69eee7319b5bf0f47bef0fcdbd8b496b3R573-R574) is a SymPy issue which I raised here: https://github.com/sympy/sympy/issues/23719. We can just leave it as-is until it gets fixed on their end.

MilesCranmer:
Thanks @tttc3, that's a great idea! I will implement it and re-upload.

MilesCranmer:
Updates:

1. I tested out the idea of putting it all in `latex`. However with the change, it seems like there was too much functionality to one function–it seemed cleaner to have a separate function. More importantly, whereas current functions (`sympy`, `latex`, `jax`, `torch`) accept only a single integer (assuming `nout_=1`), the function `latex_table` accepts a list of integers to put in the table. When `nout_>1`, they accept a list of integers, and `latex_table` will accept a list of lists. But letting `latex` work with both single-output and multi-output seems too tricky to work into the API. What do you think?
2. `latex_table()` now works for multi-output equations! They are simply in separate tables right now, though perhaps I could look at combining them in one somehow.
3. I implemented the change to include `columns` as a parameter so the user can specify the order and which columns are given.
4. Regarding sorting - this and other fine-tuning might be better left to the user? What do you think?

Cheers,
Miles

tttc3:
Re 1: I think this is a fair point, it would probably end up confusing if the methods differed too much. Also, depending on the desire to add other output formats in the future (e.g. txt, CSV, etc...), your current implementation would allow easier generalisation to other table output formats.

Re 4: I agree, sorting rows is not particularly arduous in comparison to rearranging columns.

MilesCranmer:
Great points, thanks!

By the way, do you (or @kazewong?) know if there's a way to automatically break equations **inside a table**? I found this: https://tex.stackexchange.com/a/3785/140440 but it seems to not work inside tables. Since many discovered equations are quite long, it would be nice if `latex_table` automatically broke equations into multiple lines.

kazewong:
I think with some engineering you should be able to get away with the methods shown in these two pages:
https://www.overleaf.com/learn/latex/Tables (Scroll down to the section Combining rows and columns)
https://texblog.org/2012/12/21/multi-column-and-multi-row-cells-in-latex-tables/

MilesCranmer:
Thanks!

Okay it looks like this can be done with:

```latex
\vbox{
\begin{flushleft}
$\displaystyle <equation>$
\end{flushleft}
}
```
This will automatically break it into pieces inside the table. So I guess this can be added for equations that are too long.

MilesCranmer:
Did some updates to the generated output. This is what it looks like for long equations:
```latex
\begin{table}[h]
\begin{center}
\begin{tabular}{@{}cccc@{}}
\toprule
Equation & Complexity & Loss & Score \\
\midrule
$y = 5.9560$ & $1$ & $24.388$ & $0.0$ \\
$y = 4.4687 x_{0}$ & $3$ & $4.1362$ & $0.88715$ \\
$y = 2.0248 x_{0}^{2}$ & $5$ & $0.61547$ & $0.95257$ \\
$y = 1.8967 x_{0}^{2} + 0.66315$ & $7$ & $0.41267$ & $0.19987$ \\
\begin{minipage}{0.8\linewidth} \vspace{-1em} \begin{dmath*} y = x_{0} \left(x_{0} \left(0.84477 x_{0} - 1.1845\right) + 3.2887 + \frac{0.19070 - x_{0}}{x_{0}}\right) + 0.59160 \end{dmath*} \end{minipage} & $29$ & $0.0061356$ & $1.7765 \cdot 10^{-12}$ \\
\bottomrule
\end{tabular}
\end{center}
\end{table}
```
which renders as:
image

What do you think?

tttc3:
Looks good to me. I have heard anecdotal incompatibility issues with `breqn` but have never used the package myself, perhaps @kazewong is more informed on this?

MilesCranmer:
I wonder if I should try to align the equations with equation breaks, like shown here: https://tex.stackexchange.com/a/347011/140440? I am not sure how to make a table with this, though.

kazewong:
> I wonder if I should try to align the equations with equation breaks, like shown here: https://tex.stackexchange.com/a/347011/140440? I am not sure how to make a table with this, though.

Maybe try align left instead of center? Assuming the left hand side is always just y, this should work just fine, right? more specifically, change 
`\begin{tabular}{@{}cccc@{}}`
to 
`\begin{tabular}{@{}lccc@{}}`

image

BTW, 'toprule', `midrule`, `bottomrule`, and `dmath` doesn't work out of the box on overleaf. They are probably package depencies problem. Should probably fix them

MilesCranmer:
- `toprule`, `midrule`, `bottomrule` = `\usepackage{booktabs}` (nice-looking tables)
- `dmath` = `\usepackage{breqn}` (automatic equation breaks)

I think a warning could be printed that one should include them? I'm not sure. The other option for equation breaking is `amsmath`'s `\displaymath`, but it's another dependency people will have to manually include. It also doesn't look as good.

In your example:
image
this weird alignment is what `breqn` helps avoid - see how the equation is aligned correctly even though it is broken:
image

MilesCranmer:
But, in that code, you are manually typing `\\&`. `breqn` will automatically do that for you. Is there a way to automatically do that with `amsmath`?

PR #166, opened 2022-07-17
feat(docker): Add opencontainers image-spec to `Dockerfile`

SauravMaheshkar:
This PR makes a few tiny changes to improve the overall quality of the docker image 🐳 . For reference more annotations can be found [here](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,12 @@ ARG VERSION=latest
 
 FROM --platform=$ARCH julia:$VERSION
 
+# metainformation
+LABEL org.opencontainers.image.version = "0.9.0"
+LABEL org.opencontainers.image.authors = "Miles Cranmer"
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Thanks! Is there a way to make `org.opencontainers.image.version` pull from `version.py`, so it doesn't need to be manually updated?

SauravMaheshkar:
> Thanks! Is there a way to make `org.opencontainers.image.version` pull from `version.py`, so it doesn't need to be manually updated?

Interesting suggestion, maybe we could make it to be another `ARG` that is then passed as a `--build-arg` ?

MilesCranmer:
That sounds good to me.

MilesCranmer (approved):

PR #167, opened 2022-07-25
[Feature] Easy loading from equation file

MilesCranmer:
This makes it easier to load models directly from a saved equation file. There are two ways to use this:

1. You can use `pysr.load(...)` to load from a standalone csv file of equations. However, you must pass a few different attributes here, to initialize the model: `binary_operators`, `unary_operators`, and `n_features_in`. If you have custom variable names, you also need to pass that, as well as `nout` if the number of outputs is not 1.
2. `PySRRegressor` will now automatically create a pickle file of the model parameters when running `fit(...)`. This has the same name as the equation file, but with a `.pkl` extension. This lets you run, e.g., `pysr.load("equations.csv")`, and have everything loaded from the equation. This also loads all the other parameters. (Although, you also need to pass `extra_sympy_mappings` and `extra_torch_mappings` as they are unpicklable).

@tttc3 what do you think of this?

I am also wondering if it might be more intuitive to have a `PySRRegressor.from_file(...)` constructor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.csv
 *.csv.out*
 *.bkup
+*.pkl
 performance*txt
 *.out
 trials*
--- a/README.md
+++ b/README.md
@@ -162,7 +162,15 @@ This arrow in the `pick` column indicates which equation is currently selected b
 SymPy format (`sympy_format` - which you can also get with `model.sympy()`), and even JAX and PyTorch format 
 (both of which are differentiable - which you can get with `model.jax()` and `model.pytorch()`).
 
-Note that `PySRRegressor` stores the state of the last search, and will restart from where you left off the next time you call `.fit()`. This will cause problems if significant changes are made to the search parameters (like changing the operators). You can run `model.reset()` to reset the state.
+Note that `PySRRegressor` stores the state of the last search, and will restart from where you left off the next time you call `.fit()`, assuming you have set `warm_start=True`.
+This will cause problems if significant changes are made to the search parameters (like changing the operators). You can run `model.reset()` to reset the state.
+
+You will notice that PySR will save two files: `hall_of_fame...csv` and `hall_of_fame...pkl`.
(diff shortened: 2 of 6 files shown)


MilesCranmer:
@kazewong this makes it easier to load from past runs

tttc3:
Making it simpler to load past runs is a nice addition!

I might have missed something, but I think the automatically created pickle file is of the unfitted model as it is called before `pysr._run(...)`. Is this the desired behaviour?

For me, the `from_file` constructor would be more intuitive. You can already use pickle to save and load a `PySR` instance without any additional code, ignoring the issue of the non-serializable attributes. The main feature that's being added is the ability to create a new instance of a `PySR` model from a file. This file could still be the `.pkl`, in which case, this constructor 
is mostly a convenience function for handling the reinitialization of the non-serializable attributes.

MilesCranmer:
> I might have missed something, but I think the automatically created pickle file is of the unfitted model as it is called before `pysr._run(...)`. Is this the desired behaviour?

It was the desired behavior, but your comment is making me reconsider. I guess I can see two scenarios where people would want to load their model:

1. PySR run is ongoing, or PySR run crashed, and you want to load up and visualize the output equations (which are automatically saved during the run).
2. PySR run completed successfully, and now you want to analyze the equations in a new python process.


For 1., it is necessary to checkpoint the parameters before the actual equation search finishes. With the loaded model, you can do `model.refresh()`, and it will load the checkpointed equations.

For 2., I think you could have the equations stored in the pickle file itself, and perhaps not need the additional equation file.


Maybe a nice solution for both these is to dump the pickle file twice: once before the fit runs to checkpoint the model, and once after with the fitted equations. What do you think?

tttc3:
I think dumping the pickle file twice is a good simple solution

Pablo-Lemos:
Loading is really helpful, works perfectly as far as I can tell!

MilesCranmer:
Thanks for the feedback! The merged version saves a pickle file twice - so now you can send someone that pickle file and they can evaluate your equations.

PR #175, opened 2022-08-09
Plot equation properties

Pablo-Lemos:
Similar to #156, as you suggested @MilesCranmer

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -12,7 +12,8 @@ from datetime import datetime
 import warnings
 from multiprocessing import cpu_count
 from sklearn.base import BaseEstimator, RegressorMixin, MultiOutputMixin
-from sklearn.utils import check_array, check_consistent_length, check_random_state
+from sklearn.utils import check_array, check_consistent_length, \
+    check_random_state
 from sklearn.utils.validation import (
     _check_feature_names_in,
     check_is_fitted,
@@ -29,7 +30,6 @@ from .julia_helpers import (
 from .export_numpy import CallableEquation
 from .deprecated import make_deprecated_kwargs_for_pysr_regressor
 
-
 Main = None  # TODO: Rename to more descriptive name like "julia_runtime"
 
 already_ran = False
@@ -38,8 +38,8 @@ sympy_mappings = {
     "div": lambda x, y: x / y,
     "mult": lambda x, y: x * y,
     "sqrt_abs": lambda x: sympy.sqrt(abs(x)),
-    "square": lambda x: x**2,
-    "cube": lambda x: x**3,
+    "square": lambda x: x ** 2,
+    "cube": lambda x: x ** 3,
     "plus": lambda x, y: x + y,
     "sub": lambda x, y: x - y,
     "neg": lambda x: -x,
@@ -104,7 +104,8 @@ def _process_constraints(binary_operators, unary_operators, constraints):
             # Make sure the complex expression is in the left side.
             if constraints[op][0] == -1:
                 continue
-            if constraints[op][1] == -1 or constraints[op][0] < constraints[op][1]:
+            if constraints[op][1] == -1 or constraints[op][0] < constraints[op][
+                1]:
                 constraints[op][0], constraints[op][1] = (
                     constraints[op][1],
                     constraints[op][0],
@@ -123,7 +124,8 @@ def _maybe_create_inline_operators(binary_operators, unary_operators):
             if is_user_defined_operator:
                 Main.eval(op)
                 # Cut off from the first non-alphanumeric char:
-                first_non_char = [j for j, char in enumerate(op) if char == "("][0]
+                first_non_char = \
+                [j for j, char in enumerate(op) if char == "("][0]
                 function_name = op[:first_non_char]
                 # Assert that function_name only contains
                 # alphabetical characters, numbers,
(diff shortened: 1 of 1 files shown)


pysr/sr.py
```diff
@@ -139,11 +141,11 @@ def _maybe_create_inline_operators(binary_operators, unary_operators):
 
 
 def _check_assertions(
-    X,
-    use_custom_variable_names,
-    variable_names,
-    weights,
-    y,
+        X,
```
MilesCranmer, line 144: Are these formatting changes made by your IDE? Most of these changes don't look necessary for the PR.

pysr/sr.py
```diff
@@ -1989,11 +2004,70 @@ def get_hof(self):
...
+    def plot(self, ax, threshold=0.05, fontsize=10):
+        """Create a plot of the best equations.
+
+        Parameters
+         ----------
+        threshold : float
+             The plotting threshold. Plots only equations that with a score
+             higher than threshold*max_score
+        ax : Axes
+             The axes to draw to
```
MilesCranmer, line 2016: It should automatically create a plot and axes if not passed. i.e., `ax=None` should be the default.

pysr/sr.py
```diff
@@ -1989,11 +2004,70 @@ def get_hof(self):
             return ret_outputs
         return ret_outputs[0]
 
+    def plot(self, ax, threshold=0.05, fontsize=10):
+        """Create a plot of the best equations.
+
+        Parameters
+         ----------
+        threshold : float
```
MilesCranmer, line 2012: These are out of order. Also, `fontsize` is missing.

More generally, it would be good to add `**kwargs` to the plotting function to tweak the other settings.

pysr/sr.py
```diff
@@ -1989,11 +2004,70 @@ def get_hof(self):
...
+        # do the plot:
+        for i, score in enumerate(scores_plot):
+            ax.hist(i * 5, bins=1, weights=[score],
+                    edgecolor='black', label=names_plot[i])
+
+        ax.set_xticks(5 * np.arange(len(scores_plot)))
+        ax.set_xticklabels(complexities_plot);
+
+        # label on the axis:
+        ax.set_xlabel('Complexity ', fontsize=10);
```
MilesCranmer, line 2042: This doesn't use the `fontsize` variable.

pysr/sr.py
```diff
@@ -1989,11 +2004,70 @@ def get_hof(self):
             return ret_outputs
         return ret_outputs[0]
 
+    def plot(self, ax, threshold=0.05, fontsize=10):
+        """Create a plot of the best equations.
+
+        Parameters
+         ----------
+        threshold : float
+             The plotting threshold. Plots only equations that with a score
```
MilesCranmer, line 2013: I think this metric is too restrictive. Would be better to simply plot all equations (default) or a user-passed list of indices.

MilesCranmer:
Thanks, added some comments. Also, after those - could you add a unit-test to make sure it works as expected?

pysr/sr.py
```diff
@@ -1989,11 +2004,70 @@ def get_hof(self):
             return ret_outputs
         return ret_outputs[0]
 
+    def plot(self, ax, threshold=0.05, fontsize=10):
```
MilesCranmer, line 2007: I think the function name should be more specific. There are many different types of plots you could want to make. This one is a specific view of the equations.

pysr/sr.py
Pablo-Lemos: They were, I have undone those now, sorry!

pysr/sr.py
Pablo-Lemos: I think plotting all will be too much of a mess (in most cases). I could change it to "number of equations to plot?"

pysr/sr.py
Pablo-Lemos: > Thanks, added some comments. Also, after those - could you add a unit-test to make sure it works as expected?

Will do!

pysr/sr.py
```diff
@@ -1989,6 +1989,72 @@ def get_hof(self):
...
+        plt_kwargs : dict, default=None
+            Dictionary of extra arguments to pass for plotting
+        """
+
+        assert threshold > 0, "Invalid threshold"
+        assert threshold < 1, "Invalid threshold"
+
+        if ax is None:
+            import matplotlib.pyplot as plt
+            ax = plt.gca()
```
MilesCranmer, line 2013: I think `plt.gca()` just gets the current/last axis, rather than creating a new one. But I think we would want this to generate a new figure/axis.

MilesCranmer:
Maybe call it `plot_score`? Calling it `histogram` is a bit misleading since histogram is specifically for plotting distributions, whereas this is a bar plot.

pysr/sr.py
MilesCranmer: Yeah that is a good idea! Then the default would be, say, 5 equations - the 5 with the highest score. If there are less than 5 equations, it would be equal to the number of equations. (i.e., the parameter would be default to `None` and the number inferred).

MilesCranmer:
Actually, I wonder if it would be better to allow the user to specify what column they wish to plot? i.e., `y="score"` for plotting the score of each equation or `y="loss"` for plotting the losses. I don't see a reason to have them in separate functions, so one function is probably better.

MilesCranmer:
Closing this as the upcoming tensorboard support will make this much easier to set up. Likely need to rewrite with that in mind (I will take care of this). Thanks @Pablo-Lemos!

PR #177, opened 2022-08-10
Change "best" model_selection to include a loss threshold

MilesCranmer:
In PySR `v0.10.0`, this should change the definition of `model_selection="best"` to also include a loss threshold. Sometimes, just returning the equation with the max score will have a very bad loss - perhaps because a very simple equation just happens to have a large derivative in the loss-complexity curve. 

This PR changes this so that only equations with `loss < 1.5 * min_loss` are considered. This is what was used by PySR in the GECCO contest and seemed to work well in that situation.

To select purely based on score, one can now use `model_selection="score"`

This PR also refactors the model_selection to avoid shotgun edits.


MilesCranmer:
Should add a unit-test with a manually-created dataframe so that the model selection strategies work as expected.

PR #186, opened 2022-09-06
Set JULIA_PROJECT, use Pkg.add once

mkitti:
* Sets `JULIA_PROJECT` before loading pyjulia so that PyCall.jl can be contained within the pysr environment
* Also use `Pkg.add` in a single step to add both SymbolicRegression.jl and ClusterManagers.jl to the environment at the same time

I likely advised against using the environment variable `JULIA_PROJECT` in the past. However, I think this may be necessary to avoid interference from other projects if installed within the same environment.

### DIFF
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -15,14 +15,18 @@ def install(julia_project=None, quiet=False):  # pragma: no cover
 
     # Set JULIA_PROJECT so that we install in the pysr environment
     julia_project, is_shared = _get_julia_project(julia_project)
-    os.environ["JULIA_PROJECT"] = "@" + julia_project if is_shared else julia_project
+    os.environ["JULIA_PROJECT"] = "@" + str(julia_project) if is_shared else str(julia_project)
 
     import julia
 
     julia.install(quiet=quiet)
 
+    if is_shared:
+        # is_shared is only true if the julia_project arg was None, see _get_julia_project
+        Main = init_julia(None)
+    else
+        Main = init_julia(julia_project)
 
-    Main = init_julia(julia_project)
     Main.eval("using Pkg")
 
     io = "devnull" if quiet else "stderr"
@@ -81,7 +85,7 @@ def init_julia(julia_project=None):
     from julia.core import JuliaInfo, UnsupportedPythonError
 
     julia_project, is_shared = _get_julia_project(julia_project)
-    os.environ["JULIA_PROJECT"] = "@" + julia_project if is_shared else julia_project
+    os.environ["JULIA_PROJECT"] = "@" + str(julia_project) if is_shared else str(julia_project)
 
     try:
         info = JuliaInfo.load(julia="julia")
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -12,17 +12,20 @@ def install(julia_project=None, quiet=False):  # pragma: no cover
 
     Also updates the local Julia registry.
     """
-
     # Set JULIA_PROJECT so that we install in the pysr environment
     julia_project, is_shared = _get_julia_project(julia_project)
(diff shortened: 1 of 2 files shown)


MilesCranmer:
Thanks! Just to check, this would still allow a user to set a custom `julia_project` in the PySRRegressor, right?

MilesCranmer:
Just tested it, and it seems to work. Cool!

MilesCranmer (approved):

MilesCranmer:
Btw, do you know how one could restart PyJulia? Right now (even before this PR), you can't change the environment after the first call to `.fit`, or the number of threads. Maybe there is an easy way to restart PyJulia to change these things, and load using the new environment variables.

MilesCranmer:
Thanks. Not sure you saw this message, but do know if there is a way to restart PyJulia?

Also - I just updated PyPI to 0.10.2 with this merge - so feel free to update the conda-forge PR as it will now have your JULIA_PROJECT change.

mkitti:
> Btw, do you know how one could restart PyJulia? Right now (even before this PR), you can't change the environment after the first call to `.fit`, or the number of threads. Maybe there is an easy way to restart PyJulia to change these things, and load using the new environment variables.

Not really, although you can kind of simulate it using `addprocs`, which I think you kind of figured out. I'm not sure if I understand about not being able to change the environment. Certainly you can change Julia project environments via `Pkg.activate`.

Perhaps you mean changing environment variables such as `JULIA_NUM_THREADS` to increase the number of the threads?

I recall Jeff Bezanson talking about interactive threads in the State of Julia 2022 talk:
https://youtu.be/N4h46_TCmGc?t=962 (see minute 16:00)

mkitti:
If it is not too much trouble, could you also create a  Github "release" for 0.10.1 and 0.10.2?

One future change I am contemplating is whether there should be a `PYSR_PROJECT` environment variable that could act as a default if `julia_project` is `None`. I need to think more about how that might work with distributed computing.

MilesCranmer:
> If it is not too much trouble, could you also create a Github "release" for 0.10.1 and 0.10.2?

Done!

> One future change I am contemplating is whether there should be a `PYSR_PROJECT` environment variable that could act as a default if `julia_project` is `None`. I need to think more about how that might work with distributed computing.

I think that is a good idea. For distributed computing I don't think it should be an issue, since the processes are all launched from within Julia, and then activated using the current project.

MilesCranmer:
Do you know what these errors are caused by? https://github.com/MilesCranmer/PySR/actions/runs/3006737521 It looks like they are new - not sure what from. Specifically they are only on macOS/windows on Julia <= 1.6

mkitti:
Those are likely related to my changes. Essentially PyCall has somehow failed to be installed in the current environment or we failed to change environments.

MilesCranmer:
I think it's failing at the `install` step of these tests: https://github.com/MilesCranmer/PySR/blob/d3026af02370e2b1b4cef9d4ca7c2cc4470857d5/.github/workflows/CI_large_nightly.yml#L42

mkitti:
Here's my guess at the moment. As you found out before, not all of the Julia API recognizes that `@` indicates a shared environment. The tests that are failing are on Julia versions that think `JULIA_PROJECT=@pysr-0.10.3` indicates that the Julia environment exists in a local folder called `@pysr-0.10.3` rather than `~/.julia/environments/pysr-0.10.3`. However, with successive Julia versions this improved, which is why we do not see the issue on Julia 1.7 and beyond.

Some of the implementations then install PyCall in the local folder rather that the shared environment in the depot. But when they try to PyCall they don't find it in the shared environment. This raises an `ArgumentError` similar the one I found yesterday in the conda-forge feedstock.

There is likely something else going on that accounts for the operating system dependence of the issue.

MilesCranmer:
I see, thanks. So I guess we should add a method to revert to the old `Pkg.activate` if the julia version is under 1.7?

PR #188, opened 2022-09-07
Fix install for Julia <=1.6

MilesCranmer:
If the Julia project is a shared env, the variable `JULIA_PROJECT` will only be set to `@pysr-{version}` if Julia >= 1.7. Otherwise, it will not be set, and the activation performed explicitly.

This also includes a cleaner version check which doesn't require Julia to be started.

@mkitti do you think you could take a look at this?

### DIFF
--- a/.github/workflows/CI_Windows.yml
+++ b/.github/workflows/CI_Windows.yml
@@ -26,7 +26,7 @@ jobs:
         shell: bash
     strategy:
       matrix:
-        julia-version: ['1.7.1']
+        julia-version: ['1.6', '1.7.1']
         python-version: ['3.9']
         os: [windows-2019]
     
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -32,10 +32,8 @@ def load_juliainfo():
 
 
 def _set_julia_project_env(julia_project, is_shared):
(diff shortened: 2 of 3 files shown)


mkitti:
The work around here is to give the an absolute path as the `JULIA_PROJECT`

MilesCranmer:
Any idea how to get that information in an OS-independent way, without loading Julia? I took a look in `JuliaInfo.load()` but there doesn't seem to be any relevant paths.

mkitti:
We could use `subprocess`.

```
>>> import subprocess
>>> julia_home = subprocess.run(["julia", "-e print(homedir())"], capture_output=True).stdout.decode()
>>> julia_home
'C:\\Users\\kittisopikulm'
>>> julia_env_dir = subprocess.run(["julia", "-e using Pkg; print(Pkg.envdir())"], capture_output=True).stdout.decode()
>>> julia_env_dir
'C:\\Users\\kittisopikulm\\.julia\\environments'
```

MilesCranmer:
Good idea, thanks

MilesCranmer:
Cool, it's [working](https://github.com/MilesCranmer/PySR/runs/8240109192)! Thanks @mkitti!

PR #191, opened 2022-09-10
Update backend: `x^y` instead of `abs(x)^y`, and high-precision constants

MilesCranmer:
This updates the backend to v0.12, which includes a couple of changes:
1. https://github.com/MilesCranmer/SymbolicRegression.jl/pull/123 by @johanbluecreek, which changes operators from, e.g., `abs(x)^y` to `x^y`. Expressions using these operators will now simply avoid cases where their output is NaN. This avoids letting the search exploit weird functional forms which one might wish to not use.
2. https://github.com/MilesCranmer/SymbolicRegression.jl/pull/119, which uses arbitrary-precision for constants. Now, when you use `PySRRegressor(precision=64)`, the constants will also be 64-bit precision. As a side effect, this also increases the speed of evaluation, since there is type stability.
3. Fix for when the dataset has zero variance. This allows you to fit datasets with a single row: useful if you want to find a symbolic approximation to a number (Ramanujan-style), like described on https://github.com/MilesCranmer/SymbolicRegression.jl/issues/127.
4. ~~This also sets the default parsimony to 0.0. I felt like it makes a lot of assumptions about the dataset scale to set any value for the default parsimony, so it's better to just leave the default turned off. The search will still find simple expressions; but it will not be as biased towards them, especially when the baseline loss is large.~~


MilesCranmer:
For example, here is how you can find the result of `pi/sqrt(2)` with PySR:

```python
import math

model = PySRRegressor(
    binary_operators="+ * / -".split(" "),
    unary_operators=["sqrt"],
    precision=64,
    complexity_of_constants=100,
    niterations=1000,
    weight_mutate_constant=0.0,
    should_optimize_constants=False,
    parsimony=0.001,
)

model.fit(
    [[2, math.pi, math.e]], [2.22144146908], variable_names=["_2", "_pi", "_e"]
)
```

At the end, you should be able to run: 
```python
model.sympy()
# _pi/sqrt(_2)
```

MilesCranmer:
Finding approximations to `pi` :)

```python
import math

model = PySRRegressor(
    binary_operators="+ * / -".split(" "),
    unary_operators=["sqrt", "square"],
    niterations=1000,
    # High precision:
    precision=64,
    # Favor simpler expressions:
    parsimony=0.003,
    # Disable constants:
    complexity_of_constants=100,
    # Prevent redundant computation:
    weight_mutate_constant=0.0,
    should_optimize_constants=False,
)

model.fit(
    [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]],
    [math.pi],
    variable_names=["_1", "_2", "_3", "_4", "_5", "_6", "_7", "_8", "_9", "_10"],
)
```

PR #197, opened 2022-09-22
Allow custom shared projects for `julia_project`

### LINKED ISSUE #196: [BUG] SymbolicRegressions.jl is not installed into julia_project
**Describe the bug**

Due to following lines, SymbolicRegressions.jl does not get installed into an environment indicated by the `julia_project` argument of `pysr.install`. `is_shared` is always false when `julia_project` is not `None`.

https://github.com/MilesCranmer/PySR/blob/b6ac303316bf72dbf49f29418c519e6dd2f120d9/pysr/julia_helpers.py#L89-L91

It's not clear to me why SymbolicRegressions.jl should only be installed into a shared environment. I think this may just be left over logic when it was being installed into a temporary directory. My thought is that we should remove the if statement.

MilesCranmer:
This will fix #196. If you specify an `@` in your `julia_project` string, like `julia_project="@my_project"`, then this will be designated a shared project (just as if you had specified `]activate @my_project` in the Julia REPL.

cc @mkitti

### DIFF
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -200,6 +200,7 @@ def init_julia(julia_project=None, quiet=False):
 
 
 def _add_sr_to_julia_project(Main, io_arg):
+    Main.eval("using Pkg")
     Main.sr_spec = Main.PackageSpec(
         name="SymbolicRegression",
         url="https://github.com/MilesCranmer/SymbolicRegression.jl",
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -33,7 +33,9 @@ def _get_julia_env_dir():
     # Have to manually get env dir:
     try:
         julia_env_dir_str = subprocess.run(
-            ["julia", "-e using Pkg; print(Pkg.envdir())"], capture_output=True
+            ["julia", "-e using Pkg; print(Pkg.envdir())"],
+            capture_output=True,
+            env=os.environ,
         ).stdout.decode()
     except FileNotFoundError:
         env_path = os.environ["PATH"]
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -64,17 +64,11 @@ def install(julia_project=None, quiet=False):  # pragma: no cover
 
     _version_assertion()
     # Set JULIA_PROJECT so that we install in the pysr environment
-    julia_project, is_shared = _process_julia_project(julia_project)
-    _set_julia_project_env(julia_project, is_shared)
+    processed_julia_project, is_shared = _process_julia_project(julia_project)
+    _set_julia_project_env(processed_julia_project, is_shared)
 
     julia.install(quiet=quiet)
-
-    if is_shared:
-        # is_shared is only true if the julia_project arg was None
-        # See _process_julia_project
-        Main = init_julia(None)
-    else:
-        Main = init_julia(julia_project)
+    Main = init_julia(julia_project)
 
     Main.eval("using Pkg")
 
@@ -84,7 +78,7 @@ def install(julia_project=None, quiet=False):  # pragma: no cover
     # Can't pass IO to Julia call as it evaluates to PyObject, so just directly
     # use Main.eval:
     Main.eval(
-        f'Pkg.activate("{_escape_filename(julia_project)}", shared = Bool({int(is_shared)}), {io_arg})'
+        f'Pkg.activate("{_escape_filename(processed_julia_project)}", shared = Bool({int(is_shared)}), {io_arg})'
     )
     if is_shared:
         # Install SymbolicRegression.jl:
@@ -117,14 +111,14 @@ def _import_error_string(julia_project=None):
 def _process_julia_project(julia_project):
     if julia_project is None:
         is_shared = True
-        julia_project = f"pysr-{__version__}"
+        processed_julia_project = f"pysr-{__version__}"
     elif julia_project[0] == "@":
         is_shared = True
-        julia_project = julia_project[1:]
+        processed_julia_project = julia_project[1:]
     else:
         is_shared = False
-        julia_project = Path(julia_project)
-    return julia_project, is_shared
+        processed_julia_project = Path(julia_project)
+    return processed_julia_project, is_shared
 
 
 def is_julia_version_greater_eq(juliainfo=None, version=(1, 6, 0)):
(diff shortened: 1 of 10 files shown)


test/test_env.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class TestJuliaProject(unittest.TestCase):
+    def test_custom_shared_env(self):
+        """Test that we can use PySR in a custom shared env"""
+        test_env_name = "@pysr_test_env"
+        julia_helpers.install(julia_project=test_env_name, quiet=True)
+        Main = julia_helpers.init_julia(julia_project=test_env_name)
+        Main.eval("using SymbolicRegression")
+        # TODO: Test that we are actually in the correct env.
```
MilesCranmer, line 14: @mkitti what is the best way to test that we are in a shared env here, for all operating systems/package managers?

test/test_env.py
```diff
@@ -0,0 +1,15 @@
...
+        # TODO: Delete the env at the end.
```
MilesCranmer, line 15: Not sure if there is any way to do this? I suppose I need to create a temp env instead (but that's a different question).

mkitti:
Merge https://github.com/conda-forge/pysr-feedstock/pull/51 to unpin openlibm then the test here will pass

test/test_env.py
mkitti: We could check of the absolute path of the environment starts with one of `Pkg.envdir.(DEPOT_PATH)`

test/test_env.py
mkitti: You could `pushfirst!` a temp directory into `DEPOT_PATH`. Then the shared environment would be created in the temp directory. Then you can `pop!` the tempdir off `DEPOT_PATH` and recursively delete it.

test/test_env.py
MilesCranmer: What is the most reliable way to get the absolute path of the currently-activated environment?

mkitti: `dirname(Base.active_project())`

MilesCranmer: Thanks!

test/test_env.py
MilesCranmer: If the `DEPOT_PATH` changes, and you activate a new environment, will it correctly update? It seems like it is not switching to the new setting for some reason (see error)

```python
Traceback (most recent call last):
  File "/Users/mcranmer/Documents/PySR/test/test_env.py", line 24, in test_custom_shared_env
    self.assertEqual(cur_project_dir, potential_shared_project_dirs)
AssertionError: '/Users/mcranmer/.julia/environments' != '/var/folders/1h/xyppkvx52cl6w3_h8bw_gdqh0000gr/T/tmp8dknp59s/environments'
```

i.e., the current project is still in `/Users/mcranmer/.julia/environments` rather than the new depot

MilesCranmer:
Ah, I think I see an issue (at least, for testing this). Right now the `init_julia` only sets the environment via an environment variable - rather than the `activate()` command. The only time `activate()` is used explicitly is in the install part:
https://github.com/MilesCranmer/PySR/blob/b6ac303316bf72dbf49f29418c519e6dd2f120d9/pysr/julia_helpers.py#L86-L88

So I think what we should actually do is call `activate()` if Julia has already been initialized. (Since changing the environment variable won't change the env!)

mkitti:
> So I think what we should actually do is call activate() if Julia has already been initialized. (Since changing the environment variable won't change the env!)

Correct. The only reason I used the environment variable earlier was to select the environment where we expect PyCall to be, which may not be the default environment.

MilesCranmer:
Okay so now the `DEPOT_PATH` correctly sets the project. However, for some reason, `julia.install()` switches the environment back to the normal `DEPOT_PATH`... Not sure why.

```python
from pysr.julia_helpers import init_julia, install

Main = init_julia()
Main.eval('pushfirst!(DEPOT_PATH, "/tmp/1")')
Main = init_julia("@test2")
```
This correctly goes to the `/tmp/1/environments`. However! When you do:

```python
install("@test2")
```
this sets to the normal `DEPOT_PATH`... 
```
  Activating project at `~/.julia/environments/test2`
```
Is this something in PyJulia maybe?

mkitti:
> However! When you do:
> 
> ```python
> install("@test2")
> ```
> 
> this sets to the normal `DEPOT_PATH`...
> 
> ```
>   Activating project at `~/.julia/environments/test2`
> ```
> 
> Is this something in PyJulia maybe?

In this case it seems you have not manipulated `DEPOT_PATH`, so I'm guessing that `DEPOT_PATH` is just the default value.

Note that the `DEPOT_PATH` global is specific to each Julia session. Manipulating it does not influence the corresponding environment variable `JULIA_DEPOT_PATH`. Thus if the environment variable `JULIA_DEPOT_PATH` is not set and you have not manipulated `DEPOT_PATH`, the depot will be at its default location.

pysr/julia_helpers.py
```diff
@@ -64,28 +70,13 @@ def install(julia_project=None, quiet=False):  # pragma: no cover
...
+    io_arg = _get_io_arg(quiet)
 
-    if is_shared:
-        # is_shared is only true if the julia_project arg was None
-        # See _process_julia_project
-        Main = init_julia(None)
-    else:
-        Main = init_julia(julia_project)
-
-    Main.eval("using Pkg")
```
mkitti, line 79: `PackageSpec` is not imported because you removed this line.

MilesCranmer:
Thanks - the weird thing is that those commands are from the same Python session. So the DEPOT_PATH should stick around through the `install()` call…

mkitti:
Is this happening on all versions of Julia or just particular ones, perhaps less than Julia 1.7.0?

I'm thinking about the `subprocess` and `_get_julia_env_dir` stuff that we did. The Julia process run by subprocess would have no idea about changes you may have made within the current Julia process.

pysr/julia_helpers.py
MilesCranmer: Thanks!

MilesCranmer:
Okay I think that got it. It wasn't the subprocess - it was the `JULIA_DEPOT_PATH` that was missing! Good catch.

MilesCranmer:
Cool:
```
  Activating project at `/var/folders/1h/xyppkvx52cl6w3_h8bw_gdqh0000gr/T/tmpmrkyi7cm/environments/pysr_test_env`
```


Not sure what the conda errors are though.

MilesCranmer:
Ah, the conda errors were just because the `environment.yml` wasn't updated to the new Julia 1.8.0 requirement for openlibm - have since unpinned it and seems to be happy.

mkitti:
This looks like an improvement to me. I still think we should figure out how to generalize this to non-shared environments at some point.

mkitti (approved):

PR #202, opened 2022-10-10
Fix conda forge installs

### LINKED ISSUE #201: [BUG] conda version breaking
Edit: If you are seeing issues with the conda version, try setting `update=False` when initializing `PySRRegressor`.

---

The conda-forge jobs which test `conda install -c conda-forge pysr` are currently breaking. This is even with repeat attempts: https://github.com/MilesCranmer/PySR/actions/workflows/CI_conda_forge.yml.
The error:
```
ImportError: 
    Required dependencies are not installed or built.  Run the following code in the Python REPL:
```

I find this strange, since underlying feedstock has not changed in the meantime, and it seems like the julia feedstock hasn't been updated recently either. 

FYI @mkitti @ngam. I will try to look into this a bit later today.

MilesCranmer:
This will change the default behavior of PySRRegressor to **not** update the Julia packages when `.fit` is called. i.e., this will set `update=False` by default, rather than `update=True`. I think this is more in line with what a user would expect - that once they install PySR and run `pysr.install()`, all packages will remain static.

It is also causing trouble for conda-forge, since the pysr package actually includes all dependencies when first built - if any new dependency is needed, then pysr will break. This fixes #201. cc @ngam.

To update Julia packages, in pip or in conda, you may run `pysr.install()` again, or set `update=True`.

TODO:

- [x] Create test for expected behavior when a user runs `pip install -U pysr` without running `pysr.install()` a second time.
- [x] Mention in docstring for `update` kwarg that the user should make sure their pysr version is up-to-date first, before running `install()` - so there aren't any missing packages in the depot.
- [x] Bump version number.
- [ ] Create check that the `SymbolicRegression.jl` version matches what PySR expects.

### DIFF
--- a/.github/workflows/CI_conda_forge.yml
+++ b/.github/workflows/CI_conda_forge.yml
@@ -1,6 +1,6 @@
 name: conda-forge
 
-# This CI tries the conda-forge version of PYSR
+# This CI tries the conda-forge version of PySR
 
 on:
   schedule:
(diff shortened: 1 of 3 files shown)


MilesCranmer:
There isn't a clean way to check a Julia package's version right now, so I'll leave that for later. Just added a `PACKAGE_VERSION` constant to `SymbolicRegression.jl`, so that can be used in the future.

PR #228, opened 2022-11-21
Make Julia startup options configurable; set optimize=3

MilesCranmer:
cc @mkitti 

I have `optimize=3`, but not sure if that's a safe default or not. I wouldn't expect any user to find `julia_kwargs` though, so I'm leaning towards choosing the fastest option as a default if I don't expect it to increase crashes.

### DIFF
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -171,6 +171,7 @@ def init_julia(julia_project=None, quiet=False, julia_kwargs=None):
         raise ImportError(_import_error())
 
     from julia.core import Julia
+
     Main = None
     try:
         jl = Julia(**julia_kwargs)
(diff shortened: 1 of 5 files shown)


pysr/julia_helpers.py
```diff
@@ -167,20 +172,36 @@ def init_julia(julia_project=None, quiet=False):
...
+        init_set = set(julia_kwargs_at_initialization.items())
+        new_set = set(julia_kwargs.items())
+        set_diff = new_set - init_set
+        # Remove the `compiled_modules` key, since it is not a user-specified kwarg:
+        set_diff = {k: v for k, v in set_diff if k != "compiled_modules"}
+        if len(set_diff) > 0:
+            warnings.warn(
+                "Julia has already started. The new Julia options "
+                + str(set_diff)
+                + " will be ignored."
```
MilesCranmer, line 202: @mkitti I'm thinking it might be good to do this inside PyJulia instead

MilesCranmer: (For the future)

PR #237, opened 2022-12-10
Raise warning on statically-linked Python binaries

MilesCranmer:
Time-to-first-search is very slow on statically-linked versions of Python (such as packaged with conda), as precompiled code cannot be used, so things are compiled from scratch. I think this adds some friction to the user experience, so this PR introduces a warning that recommends the user try `pyenv` if startup time is important.

When https://github.com/JuliaPy/pyjulia/issues/496 is solved, this warning is no longer needed.

See https://github.com/conda-forge/python-feedstock/issues/222 for the discussion on the conda page.

### DIFF
--- a/.github/workflows/CI_Windows.yml
+++ b/.github/workflows/CI_Windows.yml
@@ -57,9 +57,15 @@ jobs:
             python -c 'import pysr; pysr.install()'
       - name: "Run tests"
         run: python -m pysr.test main
+        env:
+          JULIA_CHECK_BOUNDS: yes
       - name: "Install Torch"
         run: pip install torch # (optional import)
       - name: "Run Torch tests"
         run: python -m pysr.test torch
+        env:
+          JULIA_CHECK_BOUNDS: yes
       - name: "Run custom env tests"
         run: python -m pysr.test env
+        env:
+          JULIA_CHECK_BOUNDS: yes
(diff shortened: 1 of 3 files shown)


mkitti:
Have you ever tried a version that uses JuliaCall / PythonCall instead of pyjulia / PyCall?

MilesCranmer:
I haven't yet, no. But I suppose it would be easy to try? The only tricky parts would be changing the project management/installation, as well as some of the lower-level PyJulia calls like `JuliaInfo`, initialization kwargs, etc.

MilesCranmer:
I wonder if it has to do with the `_LIBJULIA` variable in PyJulia being cleaned up by the python gc?
https://github.com/JuliaPy/pyjulia/blob/1e3de7bbd27312f9abd200761a0c04a03c40a23d/src/julia/libjulia.py#L90-L94

`self.api` is set to an evaluation of [get_libjulia](https://github.com/JuliaPy/pyjulia/blob/1e3de7bbd27312f9abd200761a0c04a03c40a23d/src/julia/core.py#L466), which is defined [here](https://github.com/JuliaPy/pyjulia/blob/1e3de7bbd27312f9abd200761a0c04a03c40a23d/src/julia/libjulia.py#L108), which returns a global variable `_LIBJULIA`. However, that variable is actually not declared as global in that function, but just passed when the function is first defined. I wonder if that is the source of the issue?

MilesCranmer:
Added in 49ecfff8bd4e668d9c14d335e7b737adfac49f17

PR #265, opened 2023-02-13
Update backend with constraints fix

MilesCranmer:
There was a bug noted in https://github.com/MilesCranmer/SymbolicRegression.jl/pull/171. This PR includes that fix into PySR.

This will result in modest speed improvements for default settings, and massive speed improvements when parsimony is set to zero. However, it may change behavior slightly, as now the algorithm is not allowed to explore above the maximum size, ever.


MilesCranmer:
I probably need to re-run hyperparameter tuning with this fix… https://github.com/MilesCranmer/PySR/discussions/115

Issue #238, opened 2022-12-10
Repeated CI failures on Windows


MilesCranmer:
Many of the Windows tests are now failing with various segmentation faults, which appear to be randomly triggered: 

- Nightly action: https://github.com/MilesCranmer/PySR/actions/workflows/CI_large_nightly.yml
- PR action: https://github.com/MilesCranmer/PySR/pull/237

They seem to occur more frequently on older versions of Julia, and rarely on Julia 1.8.3. Regardless, a segfault anywhere is cause for concern and should be tracked down.

The errors include:

1. Early segmentation fault (Julia 1.6.7) at first run, segfault during noise test (Julia 1.6.7 and others), as well as segfaults during warm start test.

e.g., Windows:
```
 D:\a\_temp\221410f9-8bf7-4099-901d-eb9813d86c45.sh: line 1:  1098 Segmentation fault      python -m pysr.test main
Started!
```

<details><summary>also occurs on Ubuntu sometimes:</summary>

```
signal (11): Segmentation fault
in expression starting at none:0
unknown function (ip: 0x7fd6a19bc215)
unknown function (ip: 0x7fd6a19947ff)
macro expansion at /home/runner/.julia/packages/PyCall/ygXW2/src/exception.jl:95 [inlined]
convert at /home/runner/.julia/packages/PyCall/ygXW2/src/conversions.jl:94
pyjlwrap_getattr at /home/runner/.julia/packages/PyCall/ygXW2/src/pytype.jl:378
unknown function (ip: 0x7fd68d30b1bd)
unknown function (ip: 0x7fd6a19babda)
unknown function (ip: 0x7fd6a198e9d4)
pyisinstance at /home/runner/.julia/packages/PyCall/ygXW2/src/PyCall.jl:170 [inlined]
pysequence_query at /home/runner/.julia/packages/PyCall/ygXW2/src/conversions.jl:752
pytype_query at /home/runner/.julia/packages/PyCall/ygXW2/src/conversions.jl:773
pytype_query at /home/runner/.julia/packages/PyCall/ygXW2/src/conversions.jl:806 [inlined]
convert at /home/runner/.julia/packages/PyCall/ygXW2/src/conversions.jl:831
julia_kwarg at /home/runner/.julia/packages/PyCall/ygXW2/src/callback.jl:19 [inlined]
#57 at ./none:0 [inlined]
iterate at ./generator.jl:47 [inlined]
collect_to! at ./array.jl:728
[...]

PR #268, opened 2023-02-18
Bump backend version with data race fix

MilesCranmer:
https://github.com/MilesCranmer/SymbolicRegression.jl/releases/tag/v0.15.2


MilesCranmer:
Seems like it doesn't completely solve the issues in #238, but it is certainly a potential issue, so I think this is good to solve regardless.

Issue #182, opened 2022-08-30
Сorrelation as a loss function


zachanton:
Is there any way to use Pearson or Spearman correlation as a loss function?

MilesCranmer:
Not by default - these are the available losses: https://astroautomata.com/SymbolicRegression.jl/dev/losses/. As well as any loss function that can be written as a sum over per-element losses, like `loss="myloss(x, y) = abs(x - y)"` - so long as it is non-negative.

To add a more complex loss like a correlation-based loss, you can tweak this function (in Julia - but it's quite similar to Python): https://github.com/MilesCranmer/SymbolicRegression.jl/blob/7b4fecf9e136eb7b3acf29d0669e00e772efff56/src/LossFunctions.jl#L45-L49 to compare `prediction` with `dataset.y`. Both of these are Julia vectors. 

Then, give your path to the local copy of SymbolicRegression.jl to the `julia_project` parameter in `PySRRegressor`. 

Cheers,
Miles

MilesCranmer:
#276 fixes this and allows you to do this.

PR #276, opened 2023-03-12
Allow user to specify full objective functions

MilesCranmer:
Right now, a user can only specify a custom elementwise loss function. This change allows the user to customize the entire objective function, including:

- Correlation or whole array-based losses (fixes https://github.com/MilesCranmer/PySR/issues/182, https://github.com/MilesCranmer/PySR/discussions/270, https://github.com/MilesCranmer/PySR/discussions/256).
- Objectives that include symbolic manipulation (fixes https://github.com/MilesCranmer/PySR/discussions/275). See below for a cool example.
- Objectives that have symbolic-based regularizations, like soft penalties on symbolic constraints, rather than hard penalties.

Here's a cool example that let's you fix a symbolic form of an expression, as a rational function (i.e., $P(X)/Q(X)$ for polynomials $P$ and $Q$).

```python
objective = """
function my_custom_objective(tree, dataset::Dataset{T}, options) where {T<:Real}
    # Require root node to be binary, so we can split it,
    # otherwise return a large loss:
    tree.degree != 2 && return T(10000)

    left = tree.l
    right = tree.r

    # Evaluate numerator:
    l_prediction, l_flag = eval_tree_array(left, dataset.X, options)
    !l_flag && return T(10000)

    # Evaluate denominator:
    r_prediction, r_flag = eval_tree_array(right, dataset.X, options)
    !r_flag && return T(10000)

    # Impose functional form:
    prediction = l_prediction ./ r_prediction

    return sum((prediction .- dataset.y) .^ 2) / dataset.n
end
"""

model = PySRRegressor(
    binary_operators=["*", "+", "-"],
    full_objective=objective
)
```

Note that the output equations will need to be parsed/interpreted manually, because the default printing scheme doesn't know about this custom symbolic manipulation. (The root node of `((x0 + 3.2) - 0.5)` is `-`).

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -762,7 +762,6 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         julia_kwargs=None,
         **kwargs,
     ):
-
         # Hyperparameters
         # - Model search parameters
         self.model_selection = model_selection
@@ -2123,7 +2122,6 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         equation_file_contents = copy.deepcopy(self.equation_file_contents_)
 
         for output in equation_file_contents:
-
             scores = []
             lastMSE = None
             lastComplexity = 0
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -319,9 +319,9 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         argument is constrained.
         Default is `None`.
     loss : str
-        String of Julia code specifying the loss function. Can either
-        be a loss from LossFunctions.jl, or your own loss written as a
-        function. Examples of custom written losses include:
+        String of Julia code specifying an elementwise loss function.
+        Can either be a loss from LossFunctions.jl, or your own loss
+        written as a function. Examples of custom written losses include:
         `myloss(x, y) = abs(x-y)` for non-weighted, or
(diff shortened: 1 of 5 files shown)


MilesCranmer:
Would be great if there is a way to pass custom objectives as well.

Also - since this is allowing more complex functionality to be passed, it will be important to have a way of indicating to SymbolicRegression.jl which additional functions need to be passed to worker nodes.

pysr/sr.py
```diff
@@ -1217,6 +1238,9 @@ def _validate_and_set_init_params(self):
                 "to True and `procs` to 0 will result in non-deterministic searches. "
             )
 
+        if self.loss is not None and self.full_objective is not None:
+            raise ValueError("You cannot set both `loss` and `objective`.")
```
MilesCranmer, line 1242:
```suggestion
            raise ValueError("You cannot set both `loss` and `full_objective`.")
```

MilesCranmer:
- [ ] Add example to docs

dominik-rehse:
Dear @MilesCranmer,

Thank you for implementing this awesome feature, which I saw only now!

Would this - at least in principle - allow for symbolic constraints on the functional form (e.g., $\lim_{x\to\infty} f(x)=0$)? If so, could you provide any hints on how to implement something like that? I am not really familiar with Julia and could not immediately find more information on symbolic operations in the docs of [SymbolicRegression.jl](https://github.com/MilesCranmer/SymbolicRegression.jl).

Thanks,
Dominik

MilesCranmer:
Great question! Moving to #324 (improves discoverability)

PR #281, opened 2023-03-21
Complex-valued expressions

MilesCranmer:
This enables PySR to search for complex-valued expressions. Now, if you pass complex data, it will automatically search for complex expressions.

Specifying `precision=32` (default) will use `complex64` and `precision=64` will use `complex128`.

For example:

```python
import numpy as np

X = np.random.randn(100, 3) + 1j * np.random.randn(100, 3)
y = (2 + 1j) * np.cos(X[:, 0] * (0.5 - 0.3j))

model = PySRRegressor(binary_operators=["+", "-", "*"], unary_operators=["cos"])
model.fit(X, y)
```

we can look at the best expression with:

```python
model.sympy()
```

this gives me:

```
(2.0000386 + 1.0000093*I)*cos(x0*(0.49999392 - 0.2999942*I))
```

---

The one downside of this change is I have to monkey patch scikit-learn at runtime, because their `BaseEstimator` class includes an explicit check that the input data does *not* include complex numbers. Until this issue: https://github.com/scikit-learn/scikit-learn/issues/25922 is fixed, this monkey patch is required. So, whenever you import `pysr`, know that other scikit-learn objects will also skip doing this complex number check.

---

TODO:

- [x] Add complex numbers example to docs

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -498,6 +498,8 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         What precision to use for the data. By default this is `32`
         (float32), but you can select `64` or `16` as well, giving
         you 64 or 16 bits of floating point precision, respectively.
+        If you pass complex data, the corresponding complex precision
+        will be used (i.e., `64` for complex128, `32` for complex64).
         Default is `32`.
     random_state : int, Numpy RandomState instance or None
         Pass an int for reproducible results across multiple function calls.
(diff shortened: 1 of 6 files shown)


pysr/sr.py
```diff
@@ -2007,6 +2015,18 @@ def pytorch(self, index=None):
 
     def _read_equation_file(self):
         """Read the hall of fame file created by `SymbolicRegression.jl`."""
+        regexp_im = re.compile(r"\b(\d+\.\d+)im\b")
+        regexp_im_sci = re.compile(r"\b(\d+\.\d+)[eEfF]([+-]?\d+)im\b")
+        regexp_sci = re.compile(r"\b(\d+\.\d+)[eEfF]([+-]?\d+)\b")
+        apply_regexp_im = lambda x: regexp_im.sub(r"\1j", x)
+        apply_regexp_im_sci = lambda x: regexp_im_sci.sub(r"\1e\2j", x)
+        apply_regexp_sci = lambda x: regexp_sci.sub(r"\1e\2", x)
```
MilesCranmer, line 2023: Should move these into a function.

pysr/sr.py
```diff
@@ -2329,3 +2343,16 @@ def _csv_filename_to_pkl_filename(csv_filename) -> str:
...
+_regexp_im = re.compile(r"\b(\d+\.\d+)im\b")
+_regexp_im_sci = re.compile(r"\b(\d+\.\d+)[eEfF]([+-]?\d+)im\b")
+_regexp_sci = re.compile(r"\b(\d+\.\d+)[eEfF]([+-]?\d+)\b")
+
+_apply_regexp_im = lambda x: _regexp_im.sub(r"\1j", x)
+_apply_regexp_im_sci = lambda x: _regexp_im_sci.sub(r"\1e\2j", x)
+_apply_regexp_sci = lambda x: _regexp_sci.sub(r"\1e\2", x)
+
+
+def _preprocess_julia_floats(s: str) -> str:
```
MilesCranmer, line 2357: /wtd Add docstring

MilesCranmer, line 2357: /wtd Add docstring

PR #288, opened 2023-04-01
Add paper 'Electron Transfer Rules of Minerals under Pressure…'

GCaptainNemo:
[paper] Add a paper published in Nature Communications on March 31st, which uses the `pysr` for symbolic regression.


MilesCranmer:
Awesome work!!! Thanks for submitting the paper for display on the docs too, I will merge now

Issue #257, opened 2023-01-27
[BUG] Errors when building PyCall.jl


MilesCranmer:
tldr, delete the folders:

- `~/.julia/packages/PyCall`
- `~/.julia/prefs/PyCall`
- `~/.julia/environments/pysr-{your-pysr-version}`

and re-run `python -m pysr install`.

---

I am documenting this error because it has shown up a lot for people, and I want to document the solution.

The error is basically that after upgrading Python or Julia or both, and running PySR or re-installing PySR, you get the following error:

```julia
[ Info: Trying to import PyCall...
┌ Info: PyCall is already installed but not compatible with this Python
└ executable. Re-building PyCall...
[ Info: Run `Pkg.build(“PyCall”; verbose=true)`
ERROR: LoadError: The following package names could not be resolved:
 * PyCall (not found in project or manifest)
```

The solution to this is to delete your `~/.julia/packages/PyCall`  folder. Then, simply re-run `python -m pysr install`.

If this gives you further errors in building, you should also delete `~/.julia/prefs/PyCall`, and re-run the install step. For further issues, you could delete `~/.julia/environments` folder (or more specifically, the folders `~/.julia/environments/pysr-{version}` and `~/.julia/environments/{julia_version}`, if you want to keep your other environments).

---

This should fix the problem. This issue is basically that the old installed version of PyCall.jl is attached to a specific combination of Julia and Python. And, for whatever reason, re-running `julia.install()` doesn't update the built version of PyCall.jl.

This is a temporary manual solution. In the future PySR should handle this automatically.

MilesCranmer:
It might be good to automatically reset `~/.julia/prefs/PyCall` whenever `julia.install` is called? @mkitti

MilesCranmer:
I think PySR should catch `julia.tools.PyCallInstallError` and print out some PySR-relevant information. Especially the above debugging tips.

mkitti:
Couldn't we just check for a mismatch between PyCall.python and whatever the current Python executable is? How do you find the latter?

MilesCranmer:
I'm not sure, because the Python executable could have the same name. So ideally we would want to also store the version of Python and any other meta information that might affect things.

MilesCranmer:
Also, a note to myself: should add unit tests for issues like #287 and #222. Would probably need to perform the test within a docker container with pyenv installed, and switch pyenv partway.

mkitti:
We need more diagnostics here.

[...]

PR #363, opened 2023-07-02
Get incremental installs to automatically work

MilesCranmer:
@mkitti I wonder if there is some way to do this automatically https://github.com/MilesCranmer/PySR/issues/257

TODO:

- [x] Figure out some way we can unit-test this.


MilesCranmer:
Could also do this automatically with
```python
subprocess.call(
    ["julia", "-e", 'rm(joinpath(first(DEPOT_PATH), "packages", "PyCall"), recursive=true)']
)
```
followed by re-trying `julia.install`. But I'm not sure if that is potentially dangerous or not. Wdyt @mkitti?

mkitti:
Umm, what's the primary issue. Is this new to Julia 1.9?

MilesCranmer:
This issue has been around for a while. It seems to be related to when you update your Python or Julia version, and try to re-install PySR, some install preferences stored in PyCall are violated.

mkitti:
This an environment issue:

```
ERROR: LoadError: The following package names could not be resolved: * PyCall (not found in project or manifest)
```

It is missing the UUID.

MilesCranmer:
Cool!
Screenshot 2023-07-02 at 8 09 36 PM

Automatically patches that PyCall error 🥳

pysr/test/incremental_install_simulator.dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN python3 -m pip install .
+RUN python3 -m pysr install
+
+# Change Python version:
+RUN pyenv install 3.10 && pyenv global 3.10 && pyenv uninstall -f 3.9.2
+RUN python3 -m pip install --upgrade pip
+
+# Second install of PySR:
+RUN python3 -m pip install .
+RUN rm -r ~/.julia/environments/pysr-0.14.2
```
MilesCranmer, line 51: Need to reference latest version instead

xukai92:
Hi Cranmer, I came from the https://github.com/JuliaPy/PyCall.jl/pull/945 (I was looking into a similar issue a while back).
Just wonder if you have checked https://cjdoris.github.io/PythonCall.jl/stable/juliacall/, which is an alternative to PyJulia that seems to have better Python/Julia environment/version support.

Btw, really like your works on neural-symbolic methods for physics.
I have been following them and also did [something related](https://proceedings.neurips.cc/paper/2021/hash/147540e129e096fa91700e9db6588354-Abstract.html)  :)

MilesCranmer:
Thanks and thanks for sending your exciting paper! Very cool stuff.

Yes I'm indeed interested in potentially switching to PythonCall.jl. From what I've seen the installation process is much nicer, and so it would be great to have it as the PySR -> Julia interface. Just need someone to try 🙏 🙏 . The API is pretty similar so it shouldn't be too bad to switch (basically just `jl.eval` -> `jl.seval`; and then removing `julia_helpers.py`). The only downsides are that (1) it's fairly new whereas PyJulia has been around for a while and is "battle-tested", and (2) it doesn't sound like it can eliminate the precompilation issue we've been fighting for a while with static python installations (like conda). I'm not sure if (1) is an issue in practice though because juliacall is more Julia-heavy (rather than the Python-heavy PyJulia), so it's is probably safer to use longterm and has other advantages.

Issue #362, opened 2023-07-02
[BUG]: Installation, Unsatisfiable requirements


soerenab:
### What happened?

Hi Miles, Thank you for making PySR available!

I am facing an error during installation, when calling `python3 -m pysr install`:
```julia
julia.core.JuliaError: Exception 'Unsatisfiable requirements detected for package DynamicExpressions [a40a106e]:
 DynamicExpressions [a40a106e] log:
 ├─possible versions are: 0.1.0-0.4.2 or uninstalled
 └─restricted to versions 0.8 by SymbolicRegression [8254be44] — no versions left
   └─SymbolicRegression [8254be44] log:
     ├─possible versions are: 0.18.0 or uninstalled
     └─SymbolicRegression [8254be44] is fixed to version 0.18.0' occurred while calling julia code:
Pkg.add([sr_spec, clustermanagers_spec], io=stderr)
```
I successfully installed `PySR` about two month ago following your instructions (same command as above).
From the error message, the problem seems to be that `PySR` requires `SymbolicRegression==0.18.0` which in turn requires `DynamicExpressions==0.8` which for some reason does not seem to be available.

System information:
```
Julia Version 1.8.4
Commit 00177ebc4fc (2022-12-23 21:32 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
      "Rocky Linux release 8.7 (Green Obsidian)"
```

### Version

I tried with '0.14.0' and '0.14.2'.

### Operating System

Linux

### Package Manager

pip

### Interface

IPython Terminal

### Relevant log output

```shell
python3 -m pysr install
[ Info: Julia version info
Julia Version 1.8.4
Commit 00177ebc4fc (2022-12-23 21:32 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
      "Rocky Linux release 8.7 (Green Obsidian)"
  uname: Linux 4.18.0-425.3.1.el8.x86_64 #1 SMP Wed Nov 9 20:13:27 UTC 2022 x86_64 x86_64
  CPU: Intel(R) Xeon(R) Gold 6136 CPU @ 3.00GHz:
                 speed         user         nice          sys         idle          irq
       #1-48  3000 MHz   10791459 s       2468 s    6061712 s  361330180 s     455443 s
  Memory: 377.184139251709 GB (287429.1953125 MB free)
  Uptime: 793184.29 sec
  Load Avg:  2.04  1.77  1.84
  WORD_SIZE: 64
[...]

PR #364, opened 2023-07-02
Automatically point user to installation issue page

MilesCranmer:
This should hopefully provide the debugging info up front so users don't run into #362. For every user who posts a bug report there are probably 10 who simply give up and stop using PySR... So better to make this debugging as simple as possible.


MilesCranmer:
(No clue how to unit-test this sadly... Maybe with a custom Julia depot that includes only a tiny subset of packages?)

PR #365, opened 2023-07-02
Include Python and Julia version in PySR env

MilesCranmer:
Rather than `@pysr-0.14.2` as the Julia environment, this would make it `@pysr-0.14.2-py-3.11-jl-1.9`. Thus the user would need to re-install for each combination of Python and Julia they wish to use PySR with. This is to prevent some of the issues noted in #257

cc @mkitti @ngam I wonder if this would pose an issue for conda at all? If I install a different `julia` version in conda, it would also trigger a new PySR install, right?

### DIFF
--- a/pysr/julia_helpers.py
+++ b/pysr/julia_helpers.py
@@ -124,7 +124,7 @@ def _process_julia_project(julia_project):
         )
         python_version = ".".join(map(str, sys.version_info[:2]))
         processed_julia_project = (
-            f"pysr-{__version__}" f"-py-{python_version}" f"-jl-{julia_version}"
+            f"pysr-{__version__}-py-{python_version}-jl-{julia_version}"
         )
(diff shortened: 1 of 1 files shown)


mkitti:
It is not clear to me that this will work because I'm not sure if PyCall.jl has localized the Python setting to the environment.

What we need is for PyCall.jl is to implement
https://github.com/JuliaPackaging/Preferences.jl

MilesCranmer:
Here is the PR: https://github.com/JuliaPy/PyCall.jl/pull/945

Since it’s 2 years old with no recent updates maybe the solution in the meantime is to just delete the PyCall deps whenever someone installs PySR?

MilesCranmer:
Actually it probably makes sense to include this regardless for the Julia tag, because the Manifest file might be incompatible (say some Julia package differs for older Julia versions).

Issue #388, opened 2023-07-22
FutureWarning: Index.is_object is deprecated.Use pandas.api.types.is_object_dtype instead


MilesCranmer:
Should fix this warning:
```python
FutureWarning: Index.is_object is deprecated.Use pandas.api.types.is_object_dtype instead
```

MilesCranmer:
Fixed by #389

PR #389, opened 2023-07-23
Dimensional constraints (backend version 0.21)

### LINKED ISSUE #283: [Feature] Print with sigdigits
It would be nice if the default print used, e.g., 3 significant figures, rather than all the bits in a float32. The csv file and resultant sympy equation would still be max precision, but I see no need for the progress bar to display this.

MilesCranmer:
TODO:

- [x] Allow unicode printing
- [x] Integrate units.
    - [x] We probably want to pass these as strings, otherwise we'd need to make a converter from some Python units package. Also astropy doesn't allow you to have different units in the same numpy array, so seems perhaps not optimal for this.
    - [x] X_units and y_units parameters to .fit()?
- [x] Add print_precision parameter.
- [x] Add dimensional_constraints_penalty
- [x] Consider taking advantage of v_dim_out parameter.
- [x] Reach 100% diff coverage (see todo in tests file)
- [x] Test selection mask + units
- [x] Test pickling + units
- [x] Test re-fit + units
- [x] Add parameters to param groupings
- [x] Set `y_variable_names`

Fixes #283, #388


pysr/sr.py
```diff
@@ -1361,7 +1412,10 @@ def _validate_and_set_fit_params(self, X, y, Xresampled, weights, variable_names
                     "Using DataFrame column names instead."
                 )
 
-            if X.columns.is_object() and X.columns.str.contains(" ").any():
+            if (
+                pd.api.types.is_object_dtype(X.columns)
```
MilesCranmer, line 1416: Should require certain pandas version where this function was introduced.

PR #437, opened 2023-10-03
Add paper "Discovery of a Planar Black Hole Mass Scaling Relation for Spiral Galaxies"

ZehaoJin:
I uploaded both a .gif animation and a .pdf plot. Use the plot instead of animation doesn't work.


MilesCranmer:
Hi @ZehaoJin,
Thanks so much for submitting this! Unfortunately the gif file is 10 MB which would increase the repo size by quite a bit will slow down the page loading times. Could you please only include the image? I think you can convert the pdf to an svg which will also be vectorized and small in size. (But PDFs will not easily display on the web page)
Thanks!
Miles

ZehaoJin:
Thank you very much! I put a png image instead, it should work now.

ZehaoJin:
@MilesCranmer Thank you very much! I put a png image instead, it should work now.

MilesCranmer:
Thanks. Do you know what this issue is caused by? 

https://results.pre-commit.ci/run/github/295391759/1696357984.nzf76DElTmOZ08A51kO8tw

once all the tests pass I can merge

ZehaoJin:
@MilesCranmer Thanks for pointing me to the error! Now seems like it passes

docs/papers.yml
```diff
@@ -2,6 +2,16 @@
 # information to generate the "Research Showcase"
 
 papers:
+  - title: Discovery of a Planar Black Hole Mass Scaling Relation for Spiral Galaxies
+    authors:
+      - Benjamin L. Davis (1)
+      - Zehao Jin (1)
+    affliations:
```
MilesCranmer, line 9:
```suggestion
    affliations:
```

I think this is the last remaining issue. I can fix this myself though!

MilesCranmer:
Thanks again for submitting. Very exciting result and paper by the way!! Let me know if you posted it on twitter or LinkedIn and I'm happy to help share it more widely. It is a super nice application!

PR #473, opened 2023-11-24
Added "min" and "max" sympy mapping

### LINKED ISSUE #183: [BUG] min binary function raises errors during prediction
**Describe the bug**
When I use "min" and "max" in my binary operators, fitting goes well but prediction raises an error.

**Version (please include the following information):**
- OS: ubuntu 20.04
- Julia version 1.8.0
- Python version 3.8
- PySR version 0.10.1
- Does the bug still appear with the latest version of PySR? Yes

**Configuration**
```
import numpy as np

data_x = np.random.randint(-100, 100, (200, 2))
data_y = np.minimum(data_x[:,0], 8)

model = PySRRegressor(
    model_selection="best",  # Result is mix of simplicity+accuracy
    niterations=40,
    binary_operators=[
        "min", "max"
    ],
    unary_operators=[
    ],
    extra_sympy_mappings={
    },
    # ^ Define operator for SymPy as well
    loss="loss(x, y) = (x - y)^2",
    warm_start=False,
)
print('Fitting data')
model.fit(data_x, data_y)
print(model)
print(model.get_best())

print(model.predict(data_x))

```

**Error message**
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

Happens in predict() call. The fit result is correct.

**Additional context**
I tried adding {'min': np.minimum}, or  {'min': min}, or {'min': sympy.Min} to extra_sympy_mappings. None worked.

I found a workaround : 

```
for row in data_x: 
    model.predict(row)
```
Not very satisfying though.

tanweer-mahdi:
To overcome the sympy bug while using Min/Max operators as discussed under this issue: https://github.com/MilesCranmer/PySR/issues/183

Fixes #183

### DIFF
--- a/pysr/export_sympy.py
+++ b/pysr/export_sympy.py
@@ -48,7 +48,7 @@ sympy_mappings = {
     "sign": sympy.sign,
     "gamma": sympy.gamma,
     "max": lambda x, y: sympy.Piecewise((y, x < y), (x, True)),
-    "min" : lambda x, y: sympy.Piecewise((x, x < y), (y, True)),
+    "min": lambda x, y: sympy.Piecewise((x, x < y), (y, True)),
 }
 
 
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Looks good, thanks!

PR #512, opened 2024-01-02
Wrap swap operands for binary ops

foxtran:
See https://github.com/MilesCranmer/SymbolicRegression.jl/pull/276


MilesCranmer:
@foxtran you could rebase on https://github.com/MilesCranmer/PySR/pull/515 to get the new changes

PR #535, opened 2024-01-27
(BREAKING) Rewrite Julia interface with PyJulia -> JuliaCall; other changes

MilesCranmer:
I managed to get a PythonCall.jl/JuliaCall version of PySR working!

Tagging @mkitti @cjdoris in case you are curious. Any suggestions about implementation are very appreciated :)

Detailed changes:

- (BREAKING) Changed PyJulia with JuliaCall
  - Swapping `eval` -> `seval`
  - Manually converting to `Vector` when calling SymbolicRegression.jl functions (otherwise would get passed as `PyList{Any}`; see https://github.com/JuliaPy/PythonCall.jl/issues/441)
  - Wrapped `equation_search` code with `jl.PythonCall.GC.disable()` to avoid multithreading-related segfaults (https://github.com/JuliaPy/PythonCall.jl/issues/298)
  - Manually convert `np.str_` to `str` before passing to `variable_names`, otherwise it becomes a `PyArray` and not a `String` (@cjdoris might be worth adding a workaround, it seems like PyJulia does this automatically)
- (BREAKING) The user no longer needs to run `python -m pysr install`. The install process is done by JuliaCall at import time.
  - Removed code related to `pysr.install()` and `python -m pysr install` because JuliaCall now handles this. 
  - `python -m pysr install` will not give a warning and do nothing.
- (BREAKING) Remove the feynman problems dataset. I don't think anybody used it, and doesn't seem good to have in a library itself.
- (BREAKING) Deprecated `julia_project` argument (ignored; no effect). The user now needs to set this up by customizing `juliapkg.json`.
- Dependencies are now managed by pyjuliapkg rather than the custom code we made. Simplifies things a lot!
- Rather than storing the raw julia variables in `PySRRegressor`, I am now storing a serialized version of them. This means you can now pickle the search state and warm-start the search in another Python process. It feels much safer to do it like this.
  - Not breaking! Because `self.raw_julia_state_` will actually deserialize it automatically for you
- SymbolicRegression is now available to import from PySR:

```python
from pysr import SymbolicRegression as SR
x1 = SR.Node(feature=1)  # Create expressions manually
```

- SymbolicRegression options are accessible in `model.julia_options_` (generated from a serialized format for pickle safety)
- Deprecated various kwargs to match SymbolicRegression.jl (old names will still work):
  - `ncyclesperiteration => ncycles_per_iteration`
  - `loss => elementwise_loss`
  - `full_objective => loss_function`

TODO:

- [ ] Fix remaining test issues
- [x] Decide if I want to import JuliaCall lazily (at first run) or if it gets installed from simply `import pysr`.
  - Will do at import.
- [x] Add back `pysr.install` with a warning message.
- [x] Add other unnecessary functions with simple deprecation messages.
- [x] Fixes needed:
  - [x] https://github.com/JuliaPy/PythonCall.jl/issues/451
    - Workaround implemented
  - [x] https://github.com/JuliaPy/PythonCall.jl/issues/433
    - Workaround implemented, though still broken on Julia < 1.9.0
  - [x] https://github.com/JuliaPy/PythonCall.jl/issues/232
    - Has workaround (implemented)
  - [x] https://github.com/conda-forge/staged-recipes/pull/25097
- [x] ~~Remove click dependency.~~ (Leaving for backwards compatibility)
- [x] Test custom `julia_project` with development version of SymbolicRegression.jl.
  - [x] Decide if we need to deprecate this in favor of telling the user to customize the juliapkg.json file.
  - (We will deprecate this)
- [x] Use the git version of SymbolicRegression.jl rather than registry version.
- [x] Test large distributed run on a slurm cluster.
- [x] Find other unused pieces of code and remove them.
- [x] Ensure coverage is pretty good.
- [x] Make sure we test all new code.
- [x] Set environment variable for `-O3`.
- [x] Raise warning if Julia was started without multiple threads, without optimization.
- [x] Add an integration test for restarting a warm start fit in a new Python session.
- [x] Get all tests passing
- [x] Make sure that the old Julia versions are actually being tested and PythonCall.jl isn't just installing Julia 1.10 on them (maybe with an env var?)
- [x] Update Colab notebook (and test it)
- [x] Update examples in docs to use `seval`
- [x] Make SymbolicRegression available (and save the Options)
- [x] Unittest behavior in:
  - [x]  jupyter notebook
  - [x] ~IPython~ [skipping for now...]

Post:

- [ ] Submit as an integration test to PythonCall


cjdoris:
Could you make an issue about converting `np.str_` please?

pysr/julia_helpers.py
```diff
@@ -1,284 +1,34 @@
...
-        julia_kwargs_at_initialization = julia_kwargs
+os.environ["PYTHON_JULIACALL_HANDLE_SIGNALS"] = "yes"
+os.environ["JULIA_NUM_THREADS"] = os.environ.get("JULIA_NUM_THREADS", "auto")
 
-    julia_initialized = True
-    if return_aux:
-        return Main, {"compiled_modules": using_compiled_modules}
-    return Main
+import juliapkg
+from juliacall import Main as jl
```
cjdoris, line 22: It's recommended not to use `Main` in package code, but instead to create a new module using `juliacall.newmodule`. See here: https://juliapy.github.io/PythonCall.jl/stable/juliacall/#Getting-started.

pysr/julia_helpers.py
```diff
@@ -1,284 +1,34 @@
...
+if os.environ.get("PYTHON_JULIACALL_HANDLE_SIGNALS", "yes") != "yes":
+    warnings.warn(
+        "PYTHON_JULIACALL_HANDLE_SIGNALS environment variable is set to something other than 'yes' or ''. "
+        + "You will experience segfaults if running with multithreading."
+    )
 
-    if not julia_initialized:
-        julia_kwargs_at_initialization = julia_kwargs
+os.environ["PYTHON_JULIACALL_HANDLE_SIGNALS"] = "yes"
+os.environ["JULIA_NUM_THREADS"] = os.environ.get("JULIA_NUM_THREADS", "auto")
```
cjdoris, line 19: As noted in another issue, this will have no effect if juliacall was already imported before this package. I'd recommend instead to tell the user to set these env vars themselves.

pysr/julia_helpers.py
MilesCranmer: I did this initially but it doesn’t work for a complex reason probably specific to me.

Basically it broke the distributed tests, since the extra Julia processes — created with `addprocs` from within SymbolicRegression.jl — start in `Main` and are sent all the stuff created in `Main`. Maybe there’s a way to generalize this though?

MilesCranmer: I also want to make it so that if the user starts `juliacall` and starts creating functions, all of those functions are available to pass in PySR.

Basically PySR is a top-level *tool* rather than a lightweight dependency, so I think it might be better to put it in `Main` by default..

pysr/julia_helpers.py
```diff
@@ -1,284 +1,34 @@
...
-        # use Main.eval:
-        Main.eval(
-            f'Pkg.activate("{_escape_filename(processed_julia_project)}",'
-            f"shared = Bool({int(is_shared)}), "
-            f"{io_arg})"
-        )
-
-        julia_activated_env = processed_julia_project
+# Required to avoid segfaults (https://juliapy.github.io/PythonCall.jl/dev/faq/)
+if os.environ.get("PYTHON_JULIACALL_HANDLE_SIGNALS", "yes") != "yes":
```
MilesCranmer, line 12: @cjdoris i think this might address your other comment? I didn’t do the same for threads but I can add that as well.

cjdoris: Yes I think using warnings to advise the user of the intended settings is a good idea. Setting these in the package is a bad idea in my opinion because you'll get very different behaviour depending on the order modules are imported.

Relatedly, I've added a warning in JuliaCall - if you start Julia with multiple threads but don't set `PYTHON_JULIACALL_HANDLE_SIGNALS` then you get a warning advising to set this env var.

MilesCranmer: > Setting these in the package is a bad idea in my opinion because you'll get very different behaviour depending on the order modules are imported.

I don’t disagree, but asking a user to set an environment variable to get PySR basics working is not a good user experience. Some of my users don’t even know what an environment variable is. Everything needs to be pretty automatic to attract regular scientists (my target users) — I’ve even received requests to have a downloadable app with a GUI (no code at all!).

cjdoris: Absolutely - keep it in if it's best for your users, I'm just encouraging best practice.

I wonder if it would be a good idea to be able to set these options in `juliapkg.json` to override the default behaviour. This would allow different packages to cooperate on the values, and works even when JuliaCall is already imported before your own package.

MilesCranmer: > This would allow different packages to cooperate on the values, and works even when JuliaCall is already imported before your own package.

That’s a great idea!

MilesCranmer:
Slurm test is unusually slow for some reason... I wonder if there is some type that is triggering the distributed workers to load PythonCall.jl, and thus encountering the GIL...

MilesCranmer:
@cjdoris @mkitti

https://github.com/MilesCranmer/PySR/assets/7593028/e5705ecb-9c19-4bda-b3c6-11323a29a90b


(Almost finished btw)

MilesCranmer:
So this is probably a blocker: https://github.com/JuliaPy/PythonCall.jl/issues/433

Since users can pass multi-line strings to be evaluated by Julia.

mkitti:
Could you use `begin end` blocks to turn the compound top level statement into a single top level statement?

```julia
julia> Meta.parse("""
       x = 5
       x += 3
       """)
ERROR: ParseError("extra token after end of expression")
...

julia> Meta.parse("""
       begin
           x = 5
           x += 3
       end
       """)
quote
    #= none:2 =#
    x = 5
    #= none:3 =#
    x += 3
end
```

MilesCranmer:
It turns out there’s actually a `Meta.parseall` for multiple expressions: https://github.com/JuliaPy/PythonCall.jl/pull/452

MilesCranmer:
@cjdoris what is your timeline on releasing a new version of PythonCall.jl? Just wondering if I should try to overwrite `seval` temporarily so I can get this merged.

ma-sadeghi:
@MilesCranmer Was there any trick to make Julia multiprocessing work with JuliaCall? I used to use PyJulia, and it worked fine calling a Julia function that's using Julia's multiprocessing, but I wanted to switch over to JuliaCall, but I can no longer call that function (JuliaPy/PythonCall.jl/issues/454).

I've tried a workaround, which is to use Python multiprocessing to call the serial version of that function in Julia, and that also turned out not to work either (JuliaPy/PythonCall.jl/issues/455).

I'd appreciate any pointers/tips. Thanks!

MilesCranmer:
It doesn’t seem to be an issue for me due to the way SymbolicRegression.jl uses multiprocessing — since it is basically calling `addprocs` from within the Julia code

MilesCranmer:
Still some random issues it seems... Not ready to merge yet.

```python
[ins] In [1]: from pysr import PySRRegressor, jl
ERROR: ArgumentError: Package PythonCall [6099a3de-0909-46bc-b1f4-468b9a2dfc0d] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.

Stacktrace:
  [1] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:1920
  [2] __require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1806
  [3] #invoke_in_world#3
    @ Base ./essentials.jl:921 [inlined]
  [4] invoke_in_world
    @ Base ./essentials.jl:918 [inlined]
  [5] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1797
  [6] macro expansion
    @ Base ./loading.jl:1784 [inlined]
  [7] macro expansion
    @ Base ./lock.jl:267 [inlined]
  [8] __require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1747
  [9] #invoke_in_world#3
    @ Base ./essentials.jl:921 [inlined]
 [10] invoke_in_world
    @ Base ./essentials.jl:918 [inlined]
 [11] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1740
 [12] top-level scope
    @ none:8---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 from pysr import PySRRegressor, jl

File ~/PermaDocuments/SymbolicRegressionMonorepo/.venv/lib/python3.10/site-packages/pysr/__init__.py:4
      1 # This must be imported as early as possible to prevent
      2 # library linking issues caused by numpy/pytorch/etc. importing
      3 # old libraries:
----> 4 from .julia_import import jl, SymbolicRegression  # isort:skip
      6 from . import sklearn_monkeypatch
      7 from .deprecated import best, best_callable, best_row, best_tex, install, pysr

File ~/PermaDocuments/SymbolicRegressionMonorepo/.venv/lib/python3.10/site-packages/pysr/julia_import.py:59
     53 elif os.environ["PYSR_AUTOLOAD_EXTENSIONS"] not in {"no", "yes", ""}:
     54     warnings.warn(
     55         "PYSR_AUTOLOAD_EXTENSIONS environment variable is set to something other than 'yes' or 'no' or ''."
     56     )
---> 59 from juliacall import Main as jl  # type: ignore
     62 # Finally, overwrite the seval function to use Meta.parseall
     63 # instead of Meta.parse.
     64 # TODO: Overwrite this once PythonCall.jl is updated:
     65 def seval(s: str):

File ~/PermaDocuments/SymbolicRegressionMonorepo/.venv/lib/python3.10/site-packages/juliacall/__init__.py:228
    224         raise Exception('PythonCall.jl did not start properly')
    226     CONFIG['inited'] = True
--> 228 init()
    230 def load_ipython_extension(ip):
    231     import juliacall.ipython

File ~/PermaDocuments/SymbolicRegressionMonorepo/.venv/lib/python3.10/site-packages/juliacall/__init__.py:224, in init()
    222 res = jl_eval(script.encode('utf8'))
    223 if res is None:
--> 224     raise Exception('PythonCall.jl did not start properly')
    226 CONFIG['inited'] = True

Exception: PythonCall.jl did not start properly
```

MilesCranmer:
Thanks both for your tips again! I'm very excited to get this through!! 😃

PR #559, opened 2024-03-03
Overhaul type hints with beartype and PEP 585 compliance

MilesCranmer:
This implements the following:

1. Sets up [beartype](https://github.com/beartype/beartype) runtime type checking for the entire package
    - I use the configuration that type errors => warnings. I don't want to break anything if a type is incorrect but still works; just want to let the user know a type is unexpected.
    - This runtime type checking seems pretty fast, so speed shouldn't be an issue.
    - (thanks @patrickkidger for mentioning beartype; it is lovely!)
    - BearType is not available on Python <3.8 so I only include the install on Python 3.8+
2. Updates type symbols to the PEP 585 standards.
3. Moves various stages of processing the data (like converting pandas -> numpy; getting column names; feature selection) into dataclasses so improve readability/robustness. The dataclasses have `frozen=True` so I can be sure they are not modified by other functions.

TODO:

- [ ] Only use beartype during testing rather than during runtime.
- [ ] Possibly revert `|` as shorthand for `Union` or see if there's any way to get that behavior from `future`
- [ ] Implement a `from future import __annotations__` call where the new type syntax is used


MilesCranmer:
Still not sure this should be on by default or not... Perhaps we just set it up for testing.

MilesCranmer:
(Implemented in #748)

PR #611, opened 2024-04-29
Fix sign typo in example docs

hvaara:
The discovered equation is indeed a small epsilon from the original equation, but there's a typo in the denominator in the latex simplification.


MilesCranmer:
Wow, nice catch!

PR #681, opened 2024-07-27
Enhance cross-platform compatibility for loading PySRRegressor models

zzccchen:
This PR improves the cross-platform compatibility of the `PySRRegressor.from_file` method, allowing seamless loading of models saved on different operating systems (Windows and Linux).

Key changes:

1. Introduced a custom unpickler (`CustomUnpickler`) to handle path objects from different operating systems.
2. Added a `path_to_str` function to convert path objects to strings, ensuring consistent path handling across platforms.
3. Updated the model loading process to use the custom unpickler and path conversion.
4. Maintained the existing functionality for creating models from scratch when a pickle file is not found.

These changes allow users to load PySRRegressor models on any supported platform, regardless of where the model was originally saved. This enhancement is particularly useful for users working in mixed environments or collaborating across different operating systems.

The modifications are contained within the `from_file` method to minimize impact on other parts of the codebase. The method remains backwards compatible with existing usage patterns.

Testing:
- Tested loading models saved on Linux from a Windows environment.
- Verified that the existing functionality for creating models from scratch remains intact.

This enhancement addresses the issue of cross-platform incompatibility when loading saved models, improving the overall user experience and flexibility of PySRRegressor.

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -14,11 +14,12 @@ from io import StringIO
 from multiprocessing import cpu_count
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union, cast
+from pathlib import Path, PureWindowsPath, PurePosixPath
 
 import numpy as np
 import pandas as pd
 from numpy import ndarray
-from numpy.typing import NDArray
+from numpy.typing import ArrayLike, NDArray
 from sklearn.base import BaseEstimator, MultiOutputMixin, RegressorMixin
 from sklearn.utils import check_array, check_consistent_length, check_random_state
 from sklearn.utils.validation import _check_feature_names_in  # type: ignore
@@ -949,7 +950,7 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
     @classmethod
     def from_file(
         cls,
-        equation_file: PathLike,
+        equation_file: Union[str, Path],
         *,
         binary_operators: Optional[List[str]] = None,
         unary_operators: Optional[List[str]] = None,
@@ -997,6 +998,20 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
             The model with fitted equations.
         """
 
+        class CustomUnpickler(pkl.Unpickler):
+            def find_class(self, module, name):
+                if module == 'pathlib':
+                    if name == 'PosixPath':
+                        return PurePosixPath
+                    elif name == 'WindowsPath':
+                        return PureWindowsPath
+                return super().find_class(module, name)
+
+        def path_to_str(path):
+            if isinstance(path, (PurePosixPath, PureWindowsPath)):
+                return str(path)
+            return path
+
         pkl_filename = _csv_filename_to_pkl_filename(equation_file)
 
         # Try to load model from <equation_file>.pkl
@@ -1007,11 +1022,10 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
             assert unary_operators is None
             assert n_features_in is None
             with open(pkl_filename, "rb") as f:
-                model = pkl.load(f)
-            # Change equation_file_ to be in the same dir as the pickle file
-            base_dir = os.path.dirname(pkl_filename)
-            base_equation_file = os.path.basename(model.equation_file_)
-            model.equation_file_ = os.path.join(base_dir, base_equation_file)
+                unpickler = CustomUnpickler(f)
+                model = unpickler.load()
+            # Convert equation_file_ to string to ensure cross-platform compatibility
+            model.equation_file_ = path_to_str(model.equation_file_)
 
             # Update any parameters if necessary, such as
             # extra_sympy_mappings:


MilesCranmer:
Thanks, this is great! I'll do a code review for suggesting some changes

pysr/sr.py
```diff
@@ -949,7 +949,7 @@ def __init__(
     @classmethod
     def from_file(
         cls,
-        equation_file: PathLike,
+        equation_file: Union[str, Path],
```
MilesCranmer, line 952:
```suggestion
        equation_file: PathLike,
```
`PathLike` is already equal to `Union[str, Path]` :) (It's just a convenient/easier-to-read name)

pysr/sr.py
```diff
@@ -997,6 +997,20 @@ def from_file(
...
         """
 
+        class CustomUnpickler(pkl.Unpickler):
+            def find_class(self, module, name):
+                if module == "pathlib":
+                    if name == "PosixPath":
+                        return PurePosixPath
+                    elif name == "WindowsPath":
+                        return PureWindowsPath
+                return super().find_class(module, name)
```
MilesCranmer, line 1007: Can this be moved to `utils.py`?

pysr/sr.py
```diff
@@ -997,6 +997,20 @@ def from_file(
...
+
+        def path_to_str(path):
+            if isinstance(path, (PurePosixPath, PureWindowsPath)):
+                return str(path)
+            return path
```
MilesCranmer, line 1012: Same with this – also utils.py

MilesCranmer (commented):
Thanks again, added some comments. Also could you add some unittests for `CustomUnpickler`? The test should be one that is currently failing, so this `CustomUnpickler` is required to get that unittest working.

pysr/sr.py
```diff
@@ -997,6 +997,20 @@ def from_file(
             The model with fitted equations.
         """
 
+        class CustomUnpickler(pkl.Unpickler):
```
MilesCranmer, line 1000: Can this class have a more descriptive name?

pysr/sr.py
```diff
@@ -12,13 +12,13 @@
...
 from multiprocessing import cpu_count
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union, cast
 
 import numpy as np
 import pandas as pd
 from numpy import ndarray
-from numpy.typing import NDArray
+from numpy.typing import ArrayLike, NDArray
```
MilesCranmer, line 21: Where is `ArrayLike` used?

MilesCranmer:
(By the way, feel free to click "Resolved" on any of the comments once they are completed)

pysr/sr.py
zzccchen: moved

pysr/sr.py
zzccchen: rename as `CrossPlatformPathUnpickler`

pysr/sr.py
zzccchen: removed

MilesCranmer:
Looking good! Thanks. I think it just needs some tests now – ideally tests that fail with the current master branch, but are fixed by this PR

MilesCranmer:
One option is to add a test that is only run on the Windows tests – it could generate and then pickle the model on Windows, and then create and run a dockerfile that attempts to unpickle the model?

zzccchen:
> Looking good! Thanks. I think it just needs some tests now – ideally tests that fail with the current master branch, but are fixed by this PR

Thank you for the feedback. I agree tests would be valuable, but I'm not experienced in writing test cases. Could you provide some guidance or examples for appropriate tests for these changes?

zzccchen:
> One option is to add a test that is only run on the Windows tests – it could generate and then pickle the model on Windows, and then create and run a dockerfile that attempts to unpickle the model? (Since docker runs linux).

I appreciate your suggestion, but I'm afraid I'm a bit lost on how to implement this. The process of generating, pickling, and then using Docker for unpickling sounds complex. I'm not sure where to start. Would you be able to help implement this test, or provide a more detailed guide?

MilesCranmer:
Ok what I would do is:

1. Copy https://github.com/MilesCranmer/PySR/blob/master/pysr/test/test_dev.py to `pysr/test/test_pickle.py`
2. Copy https://github.com/MilesCranmer/PySR/blob/master/pysr/test/test_dev_pysr.dockerfile to `pysr/test/test_pickle.dockerfile`
3. Modify the code https://github.com/MilesCranmer/PySR/blob/3aee19e38ceb3e0e1617d357a831400e01204658/pysr/test/test_dev.py#L8-L47 within `test_pickle.py` to run a simple PySR example where the equation file parameter set to some `Path` to a folder created with https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory. Also change the class name to `TestPickle` or something.
4. Modify https://github.com/MilesCranmer/PySR/blob/3aee19e38ceb3e0e1617d357a831400e01204658/pysr/test/test_dev.py#L51-L59 to reference `TestPickle`.
5. Search for any appearances of `test_dev` throughout the codebase, and use that to figure out where you need to add the equivalent `test_pickle`.
6. Modify the `test_pickle.dockerfile`, specifically all of these lines: https://github.com/MilesCranmer/PySR/blob/3aee19e38ceb3e0e1617d357a831400e01204658/pysr/test/test_dev_pysr.dockerfile#L34-L55 to try to load the pickle file generated by the test you created in (3).
7. Note how the code here: https://github.com/MilesCranmer/PySR/blob/3aee19e38ceb3e0e1617d357a831400e01204658/pysr/test/test_dev.py#L12-L47 builds and runs a dockerfile. Do the same in your new `test_pickle.py`, but reference the `test_pickle.dockerfile`. You should be able to pass the filename as an argument.


I think this should work. It might be overcomplicated though, maybe there's an easier way to test this. Feel free to take a different approach to test your new addition. For simple pure-Python tests you can just add a new test case to the class in `test.py`, such as https://github.com/MilesCranmer/PySR/blob/3aee19e38ceb3e0e1617d357a831400e01204658/pysr/test/test.py#L285-L297

PR #687, opened 2024-07-31
Apptainer definition file for PySR

wkharold:
fix: 686

This pull request adds an Apptainer definition file to PySR. Researchers can use it to build an Apptainer container that packages PySR and its dependencies (via a conda environment).

### DIFF
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -27,7 +27,7 @@ jobs:
         arch: ['linux/amd64']
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - uses: eWaterCycle/setup-apptainer@v2
         with:
             apptainer-version: 1.3.0
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -32,10 +32,9 @@ jobs:
             apptainer-version: 1.3.0
       - name: Build apptainer
         run: apptainer build --notest pysr.sif Apptainer.def
-      - name: Create test directory
-        run: |
-          mkdir test.d && cp pysr.sif test.d
       - name: Test apptainer
         run: |
-          cd test.d
+          TMPDIR=$(mktemp -d)
+          cp pysr.sif $TMPDIR
+          cd $TMPDIR
           apptainer test ./pysr.sif
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -34,4 +34,4 @@ jobs:
       - name: Build apptainer
         run: apptainer build pysr.sif Apptainer.def
       - name: Test apptainer
-        run: apptainer exec ./pysr.sif /bin/bash -c 'pip install pytest nbval && PYTHONPATH=/pysr python3 -m pysr test main,cli,startup'
+        run: apptainer exec ./pysr.sif /bin/bash -c 'pip install pytest nbval && PYTHONPATH=/pysr:/pysr/pysr python3 -m pysr test main,cli,startup'
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -24,7 +24,6 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        arch: ['linux/amd64']
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -35,7 +35,7 @@ jobs:
         run: apptainer build --notest pysr.sif Apptainer.def
       - name: Create test directory
         run: |
-          mkdir test.d && cp pysr.sid test.d
+          mkdir test.d && cp pysr.sif test.d
       - name: Test apptainer
         run: |
           cd test.d
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -32,6 +32,11 @@ jobs:
         with:
             apptainer-version: 1.3.0
       - name: Build apptainer
-        run: apptainer build pysr.sif Apptainer.def
+        run: apptainer build --notest pysr.sif Apptainer.def
+      - name: Create test directory
+        run: |
+          mkdir test.d && cp pysr.sid test.d
       - name: Test apptainer
-        run: apptainer exec ./pysr.sif /bin/bash -c 'PATH="/home/runner/.local/bin:$PATH" PYTHONPATH="/pysr:/pysr/pysr" pip install pytest nbval && python3 -m pysr test main,cli,startup'
+        run: |
+          cd test.d
+          apptainer test ./pysr.sif
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -34,4 +34,4 @@ jobs:
       - name: Build apptainer
         run: apptainer build pysr.sif Apptainer.def
       - name: Test apptainer
-        run: apptainer exec ./pysr.sif /bin/bash -c 'pip install pytest nbval && PYTHONPATH=/pysr:/pysr/pysr python3 -m pysr test main,cli,startup'
+        run: apptainer exec ./pysr.sif /bin/bash -c 'PATH="/home/runner/.local/bin:$PATH" PYTHONPATH="/pysr:/pysr/pysr" pip install pytest nbval && python3 -m pysr test main,cli,startup'
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -34,4 +34,4 @@ jobs:
       - name: Build apptainer
         run: apptainer build pysr.sif Apptainer.def
       - name: Test apptainer
-        run: apptainer exec ./pysr.sif /bin/bash -c 'cd /pysr; pip install pytest nbval && python3 -m pysr test main,cli,startup'
+        run: apptainer exec ./pysr.sif /bin/bash -c 'pip install pytest nbval && PYTHONPATH=/pysr python3 -m pysr test main,cli,startup'
--- a/.github/workflows/CI_apptainer.yml
+++ b/.github/workflows/CI_apptainer.yml
@@ -1,0 +1,37 @@
+name: Apptainer
+
+on:
+  push:
+    branches:
+      - 'master'
+    paths:
+      - '**'
+    tags:
+      - 'v*.*.*'
+  pull_request:
+    branches:
+      - 'master'
+    paths:
+      - '**'
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        arch: ['linux/amd64']
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: eWaterCycle/setup-apptainer@v2
+        with:
+            apptainer-version: 1.3.0
+      - name: Build apptainer
+        run: apptainer build pysr.sif Apptainer.def
+      - name: Test apptainer
+        run: apptainer exec ./pysr.sif /bin/bash -c 'cd /pysr; pip install pytest nbval && python3 -m pysr test main,cli,startup'
--- a/Apptainer.def
+++ b/Apptainer.def
@@ -47,6 +47,7 @@ Stage: runtime
 
 	# Put the Julia dependencies in /pysr/depo
 	mkdir /pysr/depot
+	mkdir /pysr/env
 
 	export JULIA_DEPOT_PATH="/pysr/depot:$JULIA_DEPOT_PATH"
 	export PYTHON_JULIAPKG_PROJECT="/pysr/depot"
--- a/Apptainer.def
+++ b/Apptainer.def
@@ -25,23 +25,22 @@ Stage: runtime
 	./pysr /pysr/pysr
 
 %post
-	cd /pysr
+	# Ensure we don't use the local pysr commands:
+	cd $(mktemp -d)
 
 	export PATH="/usr/local/julia/bin:$PATH"
 
 	# Install IPython and other useful libraries:
 	pip3 install --no-cache-dir ipython matplotlib pytest nbval
+	# Install PySR and requirements:
+	pip3 install --no-cache-dir /pysr
 
-	pip3 install --no-cache-dir -r ./requirements.txt
-
-	# Install PySR:
-	pip3 install --no-cache-dir .
-
-	# Put the Julia dependencies in /pysr/depo
+	# Put the Julia dependencies in /pysr/depot
 	mkdir /pysr/depot
-	mkdir /pysr/env
-
 	export JULIA_DEPOT_PATH="/pysr/depot"
+
+	# And set a specific environment for Julia dependencies
+	mkdir /pysr/env
 	export PYTHON_JULIAPKG_PROJECT="/pysr/env"
 
 	# Pull in all the Julia dependencies
--- a/Apptainer.def
+++ b/Apptainer.def
@@ -30,7 +30,7 @@ Stage: runtime
 	export PATH="/usr/local/julia/bin:$PATH"
 
 	# Install IPython and other useful libraries:
-	pip3 install --no-cache-dir ipython matplotlib
+	pip3 install --no-cache-dir ipython matplotlib pytest nbval
 
 	pip3 install --no-cache-dir -r ./requirements.txt
 
@@ -46,6 +46,9 @@ Stage: runtime
 
 	# Pull in all the Julia dependencies
 	python3 -c 'import pysr; pysr.load_all_packages()'
+
+%test
+	python3 -m pysr test main,cli,startup
 
 %runscript
 	# Start ipython when the container is executed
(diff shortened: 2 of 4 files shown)


MilesCranmer:
Cool, thanks!

Since we have a `Dockerfile` already, could the Apptainer file be the same as that one, but in Apptainer syntax? I would prefer to avoid having two separate images. See https://github.com/MilesCranmer/PySR/blob/master/Dockerfile

Also, would be nice to have some continuous integration tests like we have for the docker file.

wkharold:
```Since we have a Dockerfile already, could the Apptainer file be the same as that one, but in Apptainer syntax?```

I had some trouble with that approach initially so I went with `conda`. Happily, what I learned doing the `conda` version allowed me to switch back to `pip3` and make it more like the Dockerfile.  I'll work on adding an integration test that mimics the Docker one.

**Comment by MilesCranmer** [comment]
Thanks, could it also pull from the python image rather than rockylinux image? Since the base image is hosted by docker I don’t see a compelling reason for the base images to be different

File: Apptainer.def [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+	cd /opt/pysr
+
+	# Install IPython and other useful libraries:
+	pip3 install --no-cache-dir ipython matplotlib
+
+	pip3 install --no-cache-dir -r /opt/pysr/requirements.txt
+
+	# Install PySR:
+	pip3 install --no-cache-dir .
+
```
MilesCranmer, line 54:
```suggestion
        python -c "import pysr"
```
This triggers the installation of Julia libraries so should be done within the install step

Apptainer.def
```diff
@@ -0,0 +1,58 @@
...
+%post
+	dnf install -y wget python{{ PYTHON_VERSION }} python{{ PYTHON_VERSION }}-pip
+
+	cd /opt
+
+	# Download the version of Julia specified by the JULIA_RELEASE and JULIA_RELEASE_VERSION arguments
+	JULIA=julia-{{ JULIA_RELEASE }}.{{ JULIA_RELEASE_VERSION }}
+
+	wget https://julialang-s3.julialang.org/bin/linux/x64/{{ JULIA_RELEASE }}/${JULIA}-linux-x86_64.tar.gz
```
MilesCranmer, line 36: Will be done automatically by importing PySR. Alternatively we could pull from both the Julia docker image and the Python image, similar to the dockerfile

Apptainer.def
```diff
@@ -0,0 +1,58 @@
...
+
+	# Install it in /opt/julia
+	tar zxf ${JULIA}-linux-x86_64.tar.gz
```
MilesCranmer, line 39: Does apptainer work for arm64?

Apptainer.def
wkharold: Done. Note that because the container file system is read-only at runtime all the Julia stuff gets installed again, in the users $HOME/.pysr/ project, at runtime.

wkharold:
Switched to bullseye for compatibility with the Dockerfile. I was using Rocky because that's what SchedMD uses as the base for the Slurm images built for Google's Cluster Toolkit (formerly HPC Toolkit). One my goals is to be able to bind mount the necessary Slurm commands and libs into a container so that you can use the Julia ClusterManagers interface to do distributed SR from within the container. Maybe your already know how to do that with Docker?

Apptainer.def
MilesCranmer: > read-only at runtime all the Julia stuff gets installed again

Hm, I don't think this shouldn't happen. It should just use the read-only environment. Maybe one of the juliacall environment variables is not set to the apptainer dir? https://juliapy.github.io/PythonCall.jl/stable/juliacall/#julia-config

MilesCranmer: Also see https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_LOAD_PATH and https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH

MilesCranmer:
> Switched to bullseye for compatibility with the Dockerfile. I was using Rocky because that's what SchedMD uses as the base for the Slurm images built for Google's Cluster Toolkit (formerly HPC Toolkit). One my goals is to be able to bind mount the necessary Slurm commands and libs into a container so that you can use the Julia ClusterManagers interface to do distributed SR from within the container. Maybe your already know how to do that with Docker?

I see, thanks. This is a good use-case. Since I have no particular need to use bullseye, so maybe we could switch the defaults of both the Dockerfile and the Apptainer to use Rocky? You can change the default here: https://github.com/MilesCranmer/PySR/blob/3aee19e38ceb3e0e1617d357a831400e01204658/Dockerfile#L6

Apptainer.def
MilesCranmer: I guess it's tricky because Julia will also have the precompilation cache which might change depending on the user's environment variables. So ideally that should get written to a writeable directory, while the packages and registry are frozen by the image.

Apptainer.def
MilesCranmer: Got it, I see. 

One other option is that we ditch the multi-stage builds, and just let the `import pysr` command install Julia via PythonCall.jl. It should be able to get the right binaries, and also simplifies the container.

MilesCranmer:
> One my goals is to be able to bind mount the necessary Slurm commands and libs into a container so that you can use the Julia ClusterManagers interface to do distributed SR from within the container. 

I do Slurm stuff with PySR, but haven't done it from a container before. In principle its possible though.

One other thing that is needed is a function to install *all* packages, even the Julia extensions like for ClusterManagers.jl. I just made a PR https://github.com/MilesCranmer/PySR/pull/688 to add this function. So if your apptainer file calls `pysr.load_all_packages()`, all the Julia packages ever needed by PySR (including ClusterManagers.jl) will be installed.

MilesCranmer:
Ok I merged https://github.com/MilesCranmer/PySR/pull/688 which will help for your stuff, do you want to merge those changes into your branch?

wkharold:
> I see, thanks. This is a good use-case. Since I have no particular need to use bullseye, so maybe we could switch the defaults of both the Dockerfile and the Apptainer to use Rocky? You can change the default here:

It doesn't look like there are Rocky based containers for Julia on Docker Hub. I think bullseye is OK for multicore work. Transitioning to Rocky would require significant changes and testing on Slurm to ensure that it worked properly.

MilesCranmer:
Oh sorry I thought you meant you wanted to switch to Rocky. I'm happy with Bullseye too though. As long as both of the container types can have overall similar images.

MilesCranmer (commented):
For adding unit-testing, I would edit this file:

https://github.com/MilesCranmer/PySR/blob/d7e87b4a825cae2190f71319fd21534758e7647f/.github/workflows/CI_docker.yml#L24-L34

and modify the build matrix to:

```yml
    strategy:
      matrix:
        os: [ubuntu-latest]
        arch: ['linux/amd64']
        container:
            - docker
            - apptainer
```

Then you can use

```yml
      if: ${{ matrix.container == "docker" }}
```

to run any docker-only steps, and similarly for apptainer. (I'm assuming there is a way to install apptainer in github actions.)

Apptainer.def
wkharold: I've played around with this some and while you can get Julia to refer to the read-only environment using JULIA_LOAD_PATH and JULIA_DEPOT_PATH it is unhappy because the file system is read only. I'll go back and get the exact error and location in the code. I've solved this in other Apptainer contexts by using a writable overlay but I was hoping to be able to put the PySR container in a Google Artifact Registry repository so that it only had to be build once.

wkharold:
This is my first foray into Github CI testing. Would you prefer a modified CI_docker.yml file or a new CI_apptainer.yml file. My naive though is that the latter would be simpler but I could easily be missing something.

Apptainer.def
wkharold: I'm testing a JULIA_DEPOT_PATH solution that looks promising. I should have a new commit today or tomorrow.

wkharold: Fixed. Thanks for the JULIA_LOAD_PATH and JULIA_DEPOT_PATH pointers.

wkharold:
Scanning the test failures it looks like most/all of them are the result of the tests being run in a read-only directory. To avoid a `version module not found` error I cd into /pysr (which is in the container's read-only file system)  before doing anything involving `pysr`. Is there a python environment variable I could set to avoid having to do that?

MilesCranmer:
Is there a way you can mount the directory it uses for testing to a writeable directory?

.github/workflows/CI_apptainer.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: eWaterCycle/setup-apptainer@v2
+        with:
+            apptainer-version: 1.3.0
+      - name: Build apptainer
+        run: apptainer build pysr.sif Apptainer.def
+      - name: Test apptainer
+        run: apptainer exec ./pysr.sif /bin/bash -c 'pip install pytest nbval && PYTHONPATH=/pysr python3 -m pysr test main,cli,startup'
```
MilesCranmer, line 37: docker has the `-v` flag for mounting a directory. Does `apptainer exec` have something like that, so that the tests have some way to write files to? It would just be for the tests (since some of the unittests write and load files)

MilesCranmer: Also wouldn't the `pip install pytests nbval` also be writing to the image here?

.github/workflows/CI_apptainer.yml
```diff
@@ -0,0 +1,37 @@
...
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        arch: ['linux/amd64']
+
+    steps:
+      - uses: actions/checkout@v2
```
MilesCranmer, line 30:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/CI_apptainer.yml
```diff
@@ -0,0 +1,37 @@
...
+  test:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        arch: ['linux/amd64']
```
MilesCranmer, line 27:
```suggestion
```

Apptainer.def
```diff
@@ -0,0 +1,60 @@
...
+
+	pip3 install --no-cache-dir -r ./requirements.txt
+
+	# Install PySR:
+	pip3 install --no-cache-dir .
+
+	# Put the Julia dependencies in /pysr/depo
+	mkdir /pysr/depot
+
+	export JULIA_DEPOT_PATH="/pysr/depot:$JULIA_DEPOT_PATH"
```
MilesCranmer, line 51:
```suggestion
	export JULIA_DEPOT_PATH="/pysr/depot"
```
(Or is there a reason to include `JULIA_DEPOT_PATH` here? I would have expected it to be empty)

Apptainer.def
```diff
@@ -0,0 +1,60 @@
...
+	# Install IPython and other useful libraries:
+	pip3 install --no-cache-dir ipython matplotlib
+
+	pip3 install --no-cache-dir -r ./requirements.txt
+
+	# Install PySR:
+	pip3 install --no-cache-dir .
+
+	# Put the Julia dependencies in /pysr/depo
+	mkdir /pysr/depot
```
MilesCranmer, line 49:
```suggestion
	mkdir /pysr/depot
	mkdir /pysr/env
```

Apptainer.def
```diff
@@ -0,0 +1,60 @@
...
+
+	export JULIA_DEPOT_PATH="/pysr/depot:$JULIA_DEPOT_PATH"
+	export PYTHON_JULIAPKG_PROJECT="/pysr/depot"
```
MilesCranmer, line 52:
```suggestion
	export PYTHON_JULIAPKG_PROJECT="/pysr/env"
```

.github/workflows/CI_apptainer.yml
wkharold: It looks like adding PYTHONPATH=/pysr obviates the need to switch to the /pysr directory. If the directory the tests are being run from is writable adding an additional mount point should be unnecessary.

Apptainer.def
wkharold: I was just mimicking what they do in the Julia [docs](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH) they explain why there.

MilesCranmer:
I tried this on my institutional cluster to see if I could use it locally, but it failed with:

```bash
> apptainer build pysr.sif Apptainer.def
INFO:    User not listed in /etc/subuid, trying root-mapped namespace
INFO:    fakeroot command not found
INFO:    Installing some packages may fail
FATAL:   Unable to build from Apptainer.def: while parsing definition: Apptainer.def: invalid section(s) specified: arguments
```

Is the `arguments` section new on 1.3.0` maybe? My institute looks to have 1.1.9.

Apptainer.def
MilesCranmer: I think those docs are written with a user environment in mind but since this is inside a container we likely want to strip references to external files?

wkharold:
>Is the arguments section new on 1.3.0` maybe? My institute looks to have 1.1.9. Maybe we could have compatibility for 1.1.0 (or even 1.0.0) to have broader support?

It got introduced in 1.2 as part of the `--build-arg / --build-arg-file` feature.  I don't have a problem with hardcoding the Julia/Python/distro bits to make it work with 1.x.

Lots of improvements have gone in since 1.1. You should encourage your institute sys admins to upgrade.

Apptainer.def
wkharold: I just tested without the trailing $JULIA_DEPOT_PATH. Seems to work fine. I'll take it out.

wkharold:
I thought adding PYTHONPATH=/psyr had fixed the 
```
ModuleNotFoundError: No module named 'pysr.version'
```
error. I no longer see it in my local builds or tests.

MilesCranmer:
> Lots of improvements have gone in since 1.1. You should encourage your institute sys admins to upgrade.

Good point. I asked them and they are planning on it in the few month timescale. I guess there will always be some clusters using older versions, so we might as well make it compatible with older versions.

MilesCranmer:
The error about `version.py` missing indicates that it might not be getting installed to the directory we expect. Or maybe when we import pysr, it is importing from the local directory, rather than `site-packages`. Basically, `version.py` only gets created when one runs `python setup.py install`.

wkharold:
I looked in the container and `version.py` ends up in `/pysr/pysr`. I added that to `PYTHONPATH` but the test still fails. I've copied the command from the log and pasted it into my shell and it works. There must be something different about the github environment.

MilesCranmer:
The GitHub actions logs show:

```
Installing collected packages: fastjsonschema, tornado, rpds-py, pyzmq, psutil, pluggy, platformdirs, nest-asyncio, iniconfig, debugpy, coverage, comm, attrs, referencing, pytest, jupyter-core, jupyter-client, jsonschema-specifications, jsonschema, ipykernel, nbformat, nbval
  WARNING: The script debugpy is installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The scripts coverage, coverage-3.12 and coverage3 are installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The scripts py.test and pytest are installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The scripts jupyter, jupyter-migrate and jupyter-troubleshoot are installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The scripts jupyter-kernel, jupyter-kernelspec and jupyter-run are installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script jsonschema is installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The script jupyter-trust is installed in '/home/runner/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```

which is odd, I thought it would install to paths within the image? Or are there other paths that need to be set before the pip installs?

wkharold:
Apptainer mounts the home directory of its initiator by default so those messages make sense but I didn't think about adding `/home/runner/.local/bin` to PATH. I'll add that and see if it changes things.

MilesCranmer:
Oh I see. But shouldn’t it be installed inside the container rather than the user’s home directory (which thus wouldn’t be saved in the container)?

MilesCranmer:
Maybe the Python `pip install` is trying to install to the home dir rather than an env?

MilesCranmer:
Looks like we can find this with `python -m site`: https://stackoverflow.com/a/46071447

wkharold:
I looked a little closer at my local environment and realized that I don't run the tests in the PySR directory. When I did that I reproduced the failure we're seeing, ugh. So I added steps to create a new directory specifically for testing in the CI action. I also experimented with the Apptainer `%test` section of the definition file and it seemed to simplify things so added that and changed `CI_apptainer.yml` to use it.

MilesCranmer (approved):
LGTM. Thanks for accommodating all of these suggested changes and in general for the great contribution!!

MilesCranmer (approved):

PR #698, opened 2024-08-13
MPI support

MilesCranmer:
So far the distributed support of PySR has relied on ClusterManagers.jl. This PR adds MPIClusterManagers.jl (and MPI.jl) which should make PySR more compatible across clusters, since MPI.jl is standardized.

@wkharold I'd be interested to hear if this works for your cluster. You can use it with:

```python
model = PySRRegressor(multithreading=False, procs=num_nodes*num_cores, cluster_manager="mpi")
```

Note the command runs `mpirun` internally so you only need to launch the job on the head node of a slurm allocation, and it will "spread out" over the job.

### DIFF
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -6,7 +6,13 @@ First, my general tips would be to avoid using redundant operators, like how `po
 
 When running PySR, I usually do the following:
 
-I run from IPython (Jupyter Notebooks don't work as well[^1]) on the head node of a slurm cluster. Passing `cluster_manager="slurm"` will make PySR set up a run over the entire allocation. I set `procs` equal to the total number of cores over my entire allocation.
+I run from IPython (Jupyter Notebooks don't work as well[^1]) on the head node of a slurm cluster.
+Passing `cluster_manager="slurm"` will make PySR set up a run over the entire allocation (another common option is `cluster_manager="mpi"` which will use MPI).
+I set `procs` equal to the total number of cores over my entire allocation.
+
+> [!NOTE]
+> When running on a cluster, you should only launch the search from a single task on the head node, rather than starting PySR on every node simultaneously.
+> The way that ClusterManagers.jl works will automatically call the correct command to spread out the processing over the topology of nodes, such as `srun` for slurm.
 
 [^1]: Jupyter Notebooks are supported by PySR, but miss out on some useful features available in IPython and Python: the progress bar, and early stopping with "q". In Jupyter you cannot interrupt a search once it has started; you have to restart the kernel. See [this issue](https://github.com/MilesCranmer/PySR/issues/260) for updates.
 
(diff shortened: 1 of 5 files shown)


MilesCranmer:
Probably also want to allow specifying MPI options like the hosts to run on.

wkharold:
I'll give this a try today. Looks interesting. I assume PySRRegressor also supports `cluster_manager="slurm"`. Both of these approaches are interesting/valuable. Note that MPI will be quite sensitive to the network topology.

wkharold:
No MPI joy. I built the branch on a Cluster Toolkit deployed Slurm cluster. Those clusters use Open MPI and/or Intel MPI (which offers better stability/performance). It looks like PySR is using MPICH. The error is [here](https://gist.github.com/wkharold/961d5568e27a49a07c8abb56836e4f43)

wkharold:
Just doing Slurm did not succeed either. I got [this](https://gist.github.com/wkharold/5a67d65fd49f38110ced4665f08d117d) error after doing a 0.19.3 install via `pip`.

MilesCranmer:
The MPI one I'm not sure about but I think the Slurm one should* work as I've usually been able to have it working on my cluster.

Things to keep in mind: PySR will run `srun` for you, so you just need to call the script a single time. It will internally dispatch using `srun` and set up the network of workers. i.e., it's a bit different from how MPI works and you would manually launch the workers yourself.

Then, the error message you are seeing:

```
ArgumentError: Package SymbolicRegression not found in current path.
- Run `import Pkg; Pkg.add("SymbolicRegression")` to install the SymbolicRegression package.
```

This is strange because it means the workers are not activating the right environment. Do you know if the workers all have access to the same folder, across nodes? Or is it a different file system?

wkharold:
>  I think the Slurm one should* work as I've usually been able to have it working on my cluster.

The trick is to set `JULIA_PROJECT` appropriately, otherwise as you mentioned the right environment is not activated.

There are a few more things that maybe I should create issues for:
- the Google Cluster Toolkit can create partitions whose nodes are spun up on demand, the 60 timeout is too short in that case
- it would be nice to be able to add additional `srun` switches, e.g., `--ntasks-per-node`, `--cpus-per-task` etc.
- when running `pysr` from a container on a cluster (with the appropriate `--bind`s) `srun` needs to invoke julia via the container rather than directly from the file system (since it won't be there)

wkharold:
> when running pysr from a container on a cluster (with the appropriate --binds) srun needs to invoke julia via the container rather than directly from the file system (since it won't be there)

According to the docs for `Distributed.addprocs` the `exename:` keyword argument specifies the name of the Julia executable. To be able to fully containerize PySR
- the container's default action should be to run Julia, e.g., in Apptainer 
```
%runscript: 
   /opt/julia-1.10.4/bin/julia
```
- the PySR interface should allow users to specify the `exename:` keyword argument

Issue #700, opened 2024-08-20
[BUG]: PySRRegressor.from_file throws AttributeError


tbuckworth:
### What happened?

When loading a saved model with `from_file`, an AttributeError is thrown:

```
AttributeError: 'PySRRegressor' object has no attribute 'dimensionless_constants_only'
```
I am using version 0.18.4 and it is possible that the model I am loading was saved under an older version (0.17.2 maybe) - could that be the problem?

### Version

0.18.4

### Operating System

Linux

### Package Manager

pip

### Interface

Script (i.e., `python my_script.py`)

### Relevant log output

```shell
msg_model = PySRRegressor.from_file(pickle_filename, extra_torch_mappings=get_extra_torch_mappings())
Checking if logs/train/cartpole/2024-07-11__04-48-25__seed_6033/symbreg/2024-07-22__04-36-52/msg/2024-07-22__04-36-55/symb_reg.pkl exists...
Loading model from logs/train/cartpole/2024-07-11__04-48-25__seed_6033/symbreg/2024-07-22__04-36-52/msg/2024-07-22__04-36-55/symb_reg.pkl
Traceback (most recent call last):
  File "/home/titus/PycharmProjects/train-procgen-pytorch/venv/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-fec4b34b7d62>", line 1, in <module>
    msg_model = PySRRegressor.from_file(pickle_filename, extra_torch_mappings=get_extra_torch_mappings())
  File "/home/titus/PycharmProjects/train-procgen-pytorch/venv/lib/python3.8/site-packages/pysr/sr.py", line 969, in from_file
    model.set_params(**pysr_kwargs)
  File "/home/titus/PycharmProjects/train-procgen-pytorch/venv/lib/python3.8/site-packages/sklearn/base.py", line 223, in set_params
    valid_params = self.get_params(deep=True)
  File "/home/titus/PycharmProjects/train-procgen-pytorch/venv/lib/python3.8/site-packages/sklearn/base.py", line 195, in get_params
    value = getattr(self, key)
AttributeError: 'PySRRegressor' object has no attribute 'dimensionless_constants_only'
```

MilesCranmer:
> I am using version 0.18.4 and it is possible that the model I am loading was saved under an older version (0.17.2 maybe) - could that be the problem?

[...]

PR #702, opened 2024-08-21
Made from_file pickle loading more robust

tbuckworth:
Added backwards compatibility for loading models pickled under older versions.

Adds missing parameters with their default values.

See issue #700

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -1013,6 +1013,17 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
             base_equation_file = os.path.basename(model.equation_file_)
             model.equation_file_ = os.path.join(base_dir, base_equation_file)
 
+            # Get constructor parameters and default values
+            params = inspect.signature(model.__init__).parameters
+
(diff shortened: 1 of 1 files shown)


pysr/sr.py
```diff
@@ -1013,6 +1013,22 @@ def from_file(
...
+                k: v
+                for k, v in params.items()
+                if k not in model.__dict__.keys()
+                and v.name != "self"
+                and v.kind != v.VAR_KEYWORD
+            }
+
+            # Assign missing attributes
+            for k, v in missing_params.items():
+                setattr(model, k, v)
```
MilesCranmer, line 1030: Could we also have this print a warning that certain parameters are being set to defaults?

tbuckworth: I've added one

MilesCranmer (commented):
Looks good. I would like to have a unit test for this just so we can verify it works. Although I’m not immediately sure how to do that. Maybe there’s a way to have a separate Python process install an old version of PySR, run it, and then attempt to load the old model in the current code?

PR #712, opened 2024-09-04
Let sympy use log2(x) instead of log(x)/log(2)

### LINKED ISSUE #711: Let sympy use log2(x) instead of log(x)/log(2)
### Discussed in https://github.com/MilesCranmer/PySR/discussions/710

<div type='discussions-op-text'>

<sup>Originally posted by **nerai** September  3, 2024</sup>
The export to sympy currently translates log2(x) to log(x)/log(2). This is cumbersome if log2 is commonly used and makes the equations hard to read.

Sympy supports log2 natively, so I feel it would be appropriate to also translate log2 into that instead. It is perhaps possible to only fix the output in sympy, but ideally, it should be done in PySR directly.</div>

nerai:
Fixes #711 

The case of log10 is also included.

I was not immediately able to run the unit tests, but I tested it with this code:
```
import numpy as np
import sympy
from sympy.codegen.cfunctions import log2

X = np.abs(np.random.randn(500, 1)) + 2
y = np.log2(X[:, 0] ** 2) + 10 / np.log10(X[:, 0])
model = PySRRegressor(
	unary_operators=[
		"log2",
		"log10",
	],
	maxsize = 20,
)
model.fit(X, y)
print(model.equations_["sympy_format"])
```

Equation: `y = (10.003 / log10(x₀)) + log2(x₀ * x₀)`
Translated to sympy: `log2(x0*x0) + 10.003168/log10(x0)`


nerai:
> codegen looks like an internal utility? https://docs.sympy.org/latest/modules/utilities/codegen.html Is there anything simpler we can use?

I'm not familiar with sympy, so I'm not sure. The [docs](https://docs.sympy.org/latest/modules/functions/elementary.html#sympy.functions.elementary.exponential.log) indicate that the regular `log` function is translated into the paired representation, though it is not explicit:
> To get a logarithm of a different base b, use log(x, b), which is essentially short-hand for log(x)/log(b).

Grepping the sympy source code for instances of log2 did not reveal anything of note, though I may have missed it.

MilesCranmer:
I am a bit worried about using this version over just `log(x, 2)` because `codegen` looks to be pretty uncommon in user code: https://github.com/search?q=/sympy.codegen.cfunctions/&type=code. Is there anything in the sympy docs that recommend this approach over just `log(x, 2)`?

nerai:
Good point. [I asked over at sympy](https://github.com/sympy/sympy/issues/27049), and it seems it's a rare but supported function. It is documented at https://docs.sympy.org/latest/modules/codegen.html#sympy.codegen.cfunctions.log2 .

MilesCranmer:
Just checking in on this – could you add the line
```
from sympy.utilities.codegen import codegen  # noqa: F401
```

to the imports? This is needed according to the docs page.

nerai:
Can you check if that's correct? My understanding of the docs page is that it refers to the codegen callable, not the codegen namespace. The code only uses the namespace, so the import should not be not required.

MilesCranmer:
Sorry for forgetting this! I just checked and indeed it doesn't seem like it's needed – sorry for the mistake!

MilesCranmer:
Wait, I take it back, it does seem like it? You just need to import directly from the `codegen.cfunctions` for it to work:

```python
[ins] In [1]: import sympy

[ins] In [2]: sympy.codegen.cfunctions.log2(4.0)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 sympy.codegen.cfunctions.log2(4.0)

AttributeError: module 'sympy' has no attribute 'codegen'
```

compared with:

```python
[ins] In [1]: from sympy.codegen.cfunctions import log2

[ins] In [2]: log2(4.0)
Out[2]: 2.00000000000000
```


So I think the reason this currently works at all is because some dependency is doing the `from sympy.codegen.cfunctions import ...`. However, to be safe, I think we should do it here as well.

nerai:
Neat, thank you :)

Issue #656, opened 2024-06-20
[BUG]: torch export fails for expressions with constant inputs e.g. exp(2)


tbuckworth:
### What happened?

sympy2torch produces a module that fails when called if a function of a constant is present in the expression.

For example:
```
from sympy import symbols, exp
from pysr import sympy2torch
import torch

x, y = symbols("x y")

expression = exp(2)

module = sympy2torch(expression, [x, y])

X = torch.rand(100, 2).float() * 10

torch_out = module(X)
```
produces this error
> TypeError: exp(): argument 'input' (position 1) must be Tensor, not float

I've tried other expressions like log(4), which produces the same problem.

The current mapping in `export_torch.py` is `sympy.exp: torch.exp`.

I believe that 

```
def exp(x):
    return torch.exp(torch.FloatTensor(x))
```
then using the mapping `sympy.exp: exp` might work, but I have been unable to test it (adding to extra_sympy_mappings doesn't work, I think because it is chained to the end of the existing mappings and doesn't override the original one).

Alternatively, perhaps simplifying all expressions to constants where possible might solve the problem for all expressions e.g. `exp(2)` becomes `7.38905609893`.

### Version

0.18.4

### Operating System

Linux

### Package Manager

pip

### Interface

Script (i.e., `python my_script.py`)

MilesCranmer:
Thanks for the bug report! One thing I am surprised about is that I thought this would already happen? See this code:
https://github.com/MilesCranmer/PySR/blob/06ca0e376e63d563aa063028a5f9bc7fa7d849c5/pysr/export_torch.py#L94-L97

The code `torch.tensor(float(expr))` should already map any constant into a torch tensor. 

Maybe the issue is that you are explicitly passing a Python integer, rather than a SymPy integer? 

For example, we can see these are actually different classes:

```python
In [4]: isinstance(1, sympy.Integer)
Out[4]: False
```

Did you see this error from a PySR export, or are you trying to use `sympy2torch` manually and putting in the integers explicitly?

tbuckworth:
This came about using `PySRRegressor.fit()`, which produced an expression containing `square(exp(sign(0.44796443)))`, which seems to simplify to `exp(2)`.

I recreated the issue using `expression = exp(sign(0.44796443))*exp(sign(0.44796443))` originally, but wrote `exp(2)` here as a minimal example.

MilesCranmer:
[...]

PR #726, opened 2024-09-26
converting sympy.NumberSymbol to torch.tensor in export_torch.py

tbuckworth:
attempting to address #656

### DIFF
--- a/pysr/export_torch.py
+++ b/pysr/export_torch.py
@@ -116,7 +116,7 @@ def _initialize_torch():
                     self._args = ()
                 elif issubclass(expr.func, sympy.NumberSymbol):
                     # Can get here from exp(1) or exact pi
-                    self._value = float(expr)
+                    self._value = torch.tensor(float(expr))
                     self._torch_func = lambda: self._value
                     self._args = ()
                 elif issubclass(expr.func, sympy.Symbol):


MilesCranmer:
Nice! Do you want to add a unit test for the MWE you described in the issue?

tbuckworth:
Hi I've added a unit test to here, let me know if that is sufficient or if i need to do anything else

MilesCranmer:
Seems like the test is failing

tbuckworth:
Apologies, I won't be able to address it until next week

MilesCranmer:
No worries!

tbuckworth:
I figured out what's going on.

`sin(sign(-0.04))` gets simplified to `-sin(1)`

`sin(1)` requires converting the argument `1`, which is a `sympy.core.numbers.One` instance, this is a subclass of `sympy.Rational`, so it is picked up by that line and converted into `float(1)`.

But `torch.sin` requires `torch.tensor` as input.

So i have added a line to make `sympy.core.numbers.One` get treated the same way as `sympy.Float` subclasses.
All the torch tests pass on my machine, so I think this is fine behaviour.

perhaps you would rather fix it at a different level of abstraction?

MilesCranmer (commented):
I wonder if the underlying issue is the use of `issubclass` instead of `isinstance`... Maybe try replacing those conditions with the following code?

```python
def __init__(self, *, expr, _memodict, _func_lookup, **kwargs):
    super().__init__(**kwargs)

    self._sympy_func = expr.func

    if isinstance(expr, sympy.Float):
        self._value = torch.nn.Parameter(torch.tensor(float(expr)))
        self._torch_func = lambda: self._value
        self._args = ()
    elif isinstance(expr, sympy.Rational) and not isinstance(expr, sympy.Integer):
        # This is some fraction fixed in the operator.
        self._value = float(expr)
        self._torch_func = lambda: self._value
        self._args = ()
    elif isinstance(expr, sympy.UnevaluatedExpr):
        if len(expr.args) != 1 or not isinstance(expr.args[0], sympy.Float):
            raise ValueError(
                "UnevaluatedExpr should only be used to wrap floats."
            )
        self.register_buffer("_value", torch.tensor(float(expr.args[0])))
        self._torch_func = lambda: self._value
        self._args = ()
    elif isinstance(expr, sympy.Integer):
        # Handles Integer special cases like NegativeOne, One, Zero
        self._value = int(expr)
        self._torch_func = lambda: self._value
        self._args = ()
    elif isinstance(expr, sympy.NumberSymbol):
        # Handles mathematical constants like pi, E
        self._value = float(expr)
        self._torch_func = lambda: self._value
        self._args = ()
    elif isinstance(expr, sympy.Symbol):
        self._name = expr.name
        self._torch_func = lambda value: value
        self._args = ((lambda memodict: memodict[expr.name]),)
    else:
        try:
            self._torch_func = _func_lookup[expr.func]
        except KeyError:
            raise KeyError(
                f"Function {expr.func} was not found in Torch function mappings. "
                "Please add it to extra_torch_mappings in the format, e.g., "
                "{sympy.sqrt: torch.sqrt}."
            )
        args = []
        for arg in expr.args:
            try:
                arg_ = _memodict[arg]
            except KeyError:
                arg_ = type(self)(
                    expr=arg,
                    _memodict=_memodict,
                    _func_lookup=_func_lookup,
                    **kwargs,
                )
                _memodict[arg] = arg_
            args.append(arg_)
        self._args = torch.nn.ModuleList(args)
```

tbuckworth:
Thanks!

That code gives the following error

> TypeError: sin(): argument 'input' (position 1) must be Tensor, not int

note that it is now saying `int` instead of `float`.

that's due to this code:

```python
elif isinstance(expr, sympy.Integer):
    # Handles Integer special cases like NegativeOne, One, Zero
    self._value = int(expr)
    self._torch_func = lambda: self._value
    self._args = ()
```

is there a reason that `int`s need to be treated this way?

for the `sin(1)` case, `torch.sin` requires that `1` is a `torch.tensor`.

would it be ok to just add the following code before the integer case?

```python
elif isinstance(expr, sympy.core.numbers.One):
    # Handles Integer special cases like NegativeOne, One, Zero
    self._value = torch.tensor(int(expr))
    self._torch_func = lambda: self._value
    self._args = ()
```

or should it be `torch.nn.Parameter(torch.tensor(float(expr)))` so that it can be fine-tuned?

either way, it then passes the test

PR #765, opened 2024-12-05
Fall back to `eager` registry when needed

DilumAluthge:


DilumAluthge:
@MilesCranmer If it fails with `eager`, after we set the environment variable to `conservative`, we need to make sure that `juliacall` does another `Pkg.Registry.update()`, otherwise the `conservative` value won't have any effect.

Does `juliacall` automatically run `Pkg.Registry.update()`, or do we need to do so manually ourselves in this package?

DilumAluthge:
> Does `juliacall` automatically run `Pkg.Registry.update()`

I'm guessing that the answer to this question is yes, based on this line (line 334):

https://github.com/JuliaPy/pyjuliapkg/blob/3a2c66019f098c1ebf84f933a46e7ca70e82792b/src/juliapkg/deps.py#L334-L334

But I'm not very familiar with `pyjuliapkg`.

pysr/julia_import.py
```diff
@@ -42,10 +59,62 @@
...
+        # So, in this case, we just rethrow the caught exception:
+        raise  # rethrow the caught exception
+    else:
+        # In this case, the user had not set JULIA_PKG_SERVER_REGISTRY_PREFERENCE.
+        # So we had initially tried with JULIA_PKG_SERVER_REGISTRY_PREFERENCE=eager, but that
+        # resulted in an exception.
+        # So let us now try with JULIA_PKG_SERVER_REGISTRY_PREFERENCE=conservative
+        os.environ["JULIA_PKG_SERVER_REGISTRY_PREFERENCE"] = "conservative"
+        try:
+            import juliacall
```
DilumAluthge, line 100:
```suggestion
            # Note: after changing the value of `JULIA_PKG_SERVER_REGISTRY_PREFERENCE`,
            # you need to run `Pkg.Registry.update()` again. Otherwise the change will not take effect.
            # `juliacall` will automatically do this for us, so we don't need to do it ourselves.
            #
            # See line 334 here:
            # https://github.com/JuliaPy/pyjuliapkg/blob/3a2c66019f098c1ebf84f933a46e7ca70e82792b/src/juliapkg/deps.py#L334-L334
            import juliacall
```

DilumAluthge:
I realize that we need to do this in every location where a Pkg operation might be needed.

DilumAluthge:
So I'll have to refactor this functionality out, to make it easier to call in multiple locations.

DilumAluthge:
@MilesCranmer I think this is ready for review now.

pysr/julia_import.py
```diff
@@ -4,6 +4,15 @@
 from types import ModuleType
 from typing import cast
 
+from .julia_registry_helper import with_juliaregistrypref
+
+
+def import_juliacall():
+    import juliacall
```
DilumAluthge, line 11:
```suggestion
    import juliacall # type: ignore
```

DilumAluthge:
@MilesCranmer Can you enable CI on the new commit?

MilesCranmer (commented):
I made a few changes. Importantly I changed it so that it doesn't change the registry preference by default. Only if it fails to resolve do I switch things to `eager`.

What do you think of the edits?

MilesCranmer:
Also, added some unittests

pysr/julia_registry_helpers.py
```diff
@@ -0,0 +1,46 @@
...
+            + f"General registry from the Julia Pkg server (via the `{PREFERENCE_KEY}` environment variable)."
+        )
+        os.environ[PREFERENCE_KEY] = "eager"
+        try:
+            return f(*args, **kwargs)
+        finally:
+            if old_value is not None:
+                os.environ[PREFERENCE_KEY] = old_value
+            else:
+                del os.environ[PREFERENCE_KEY]
```
MilesCranmer, line 46: I feel like this whole thing should go into Pkg.jl... But I suppose this behavior is needed most for Julia+Python packages so PySR is probably good to start with

MilesCranmer:
Thanks @DilumAluthge!

PR #813, opened 2025-01-21
Update papers.yml with https://doi.org/10.1051/0004-6361/202348361

sambeckers:
[before_after_pysr.pdf](https://github.com/user-attachments/files/18488684/before_after_pysr.pdf)
Added my 2024 A&A paper

### DIFF
--- a/docs/papers.yml
+++ b/docs/papers.yml
@@ -2,6 +2,20 @@
 # information to generate the "Research Showcase"
 
 papers:
+  - title: Individual chaotic behaviour of the S-stars in the Galactic centre
+    authors:
+      - Sam J. Beckers (1)
+      - Colin M. Poppelaars (1)
+      - Veronica S. Ulibarrena (1)
+      - Tjarda N. Boekholt (2)
+      - Simon F. Portegies Zwart (1)
+    affiliations:
+      1: Leiden Observatory, Leiden University
+      2: NASA Ames Research Center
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Super interesting research, nice work!

MilesCranmer:
For the figure to display, do you think you could add a subtitle (before/after)? Also is there supposed to be a legend? I can't find it on the PDF

sambeckers:
> For the figure to display, do you think you could add a subtitle (before/after)? Also is there supposed to be a legend? I can't find it on the PDF

Yes, of course! We opted to not include the legend in the paper since its more about the shift as a whole between before and after. However, I do also think for this case it's better to include it, so I added it back here.
[before_after_pysr_v2.pdf](https://github.com/user-attachments/files/18496151/before_after_pysr_v2.pdf)

sambeckers:
Any update? :)

MilesCranmer:
Hey @sambeckers,

Sorry for dropping this! 

Can you add a figure .png or .jpeg to the upload show above the paper? See https://ai.damtp.cam.ac.uk/pysr/papers for examples.

Cheers,
Miles

MilesCranmer:
Congrats on the paper! Confirming it has been merged in https://github.com/MilesCranmer/PySR/pull/1202. Apologies for the delay.

PR #920, opened 2025-05-13
Deprecate ParametricExpressionSpec in favor of TemplateExpressionSpec

MilesCranmer:
This tries to give a helpful warning message to the user for how to upgrade to template expressions.

Suggested by @gm89uk as he found the performance was much better. So there's not much of a point in having a separate class anymore when template expressions are just better.

Care to take a look at the PR and let me know what you think?

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -560,37 +560,25 @@ model = PySRRegressor(
 model.fit(X, y)
 ```
 
-You can also use parameters in your template expressions, which will be optimized during the search:
-
-```python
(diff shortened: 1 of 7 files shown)


gm89uk:
Thanks Miles. Apologies, my entire set up is now directly in Julia, I'll struggle to get this tested in a timely manner. Are you planning a similar warning in SR.jl?

MilesCranmer:
No worries! And yes; will put it in Julia too with a deprecation warning there. Just happened to write it on the Python side first

PR #929, opened 2025-05-19
ci: update workflows to test 3.13

MilesCranmer:


MilesCranmer:
Looks like we are stuck until https://github.com/TheKevJames/coveralls-python/pull/542 merges

PR #946, opened 2025-05-28
Add Sherif Samy's FRP bolted joints paper to PySR research showcase

19sSs19:
This PR adds our Scientific Reports paper to the PySR research showcase. The study combines FEM, experimental testing, and symbolic regression using PySR to derive an interpretable model for predicting damage initiation in hybrid GFRP-steel bolted joints. Thank you for the opportunity to share this work!


19sSs19:
frp

PR #1185, opened 2026-04-18
fix: keep release-please manifest on released prerelease

MilesCranmerBot:
This fixes the release-please/version ownership mismatch behind #1165.

Summary:
- stop backend update PRs from mutating `pyproject.toml`
- stop backend update PRs from mutating `.release-please-manifest.json`
- reset the release-please manifest to the last released prerelease, `2.0.0a1`

This leaves release-please as the owner of release state, so the next release PR can advance PySR to `2.0.0a2` instead of incorrectly jumping to `2.1.0`.


MilesCranmer:
@MilesCranmerBot please fix the pre-commit.ci

Also, I am confused about something. Don't we want to _not_ update the changelog, since a fixed release-please CI run would do that? i.e., the 2.0.0a2 has not been released yet, correct?

MilesCranmer:
@MilesCranmerBot does that mean we need to downgrade the .release-please manifest? I'm confused why it is already a2 if there hasn't been a GitHub release yet...

MilesCranmer:
@MilesCranmerBot but the way release-please works is that a PR gets created which both updates the manifest _and_ makes the release. That's why. But not sure.

MilesCranmer (approved):

Issue #1018, opened 2025-08-28
[Feature]: Smart default for parallelism


MilesCranmer:
### Feature Request

It would be nice to have parallelism be set in a smart way, rather than always set to `"multithreading"`. For example:

1. `deterministic == True` => `parallelism="serial"`
    - We could also forward the global numpy random seed, if the user does not provide a random state?
2. `numprocs is not None` => `parallelism="multiprocessing"`
3. `cluster_manager is not None` => `parallelism="multiprocessing"`

And in also other cases, we default to `"multithreading"`.

Issue #1001, opened 2025-07-30
[Feature]: automated tips


MilesCranmer:
### Feature Request

It might be useful to go through the discussions forum, and see if there are instances where we can automatically print “tips” to help the user. For example, for very large dataset size, we could tell the user they probably don’t need it.

(And a parameter like `tips` could turn it off)

PR #1192, opened 2026-05-19
feat: warn users when running with default low `niterations`

### LINKED ISSUE #603: [Feature]: Warn user about low `niterations`
### Feature Request

I see a lot of people using the default niterations (purposefully small) for “production runs” and not getting as good results. Either the default `niterations` should be higher, or, unless set explicitly, should raise a warning and point the user to the tuning page.

mvanhorn:
## Summary

Single source-file edit plus one new test.

### Detect "user passed niterations vs. used the default"

`PySRRegressor.__init__` currently does `niterations: int = 100,` and then `self.niterations = niterations` (`pysr/sr.py` lines 956, 1070). There is no way after the fact to tell whether the value came from the user or from the default.

Add a sentinel:

```python
_DEFAULT_NITERATIONS = 100  # module-level constant, near the top of pysr/sr.py

class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
    ...
    def __init__(
        self,
        ...,
        niterations: int = _DEFAULT_NITERATIONS,
        ...
    ):
        ...
        self.niterations = niterations
        self._niterations_was_default = niterations == _DEFAULT_NITERATIONS
```

The `__init__`-time equality check is the lightest-touch way to detect default usage. It has one edge case: a user who explicitly passes `niterations=100` is indistinguishable from the default. That is acceptable - the warning is advisory, and asking a user to set `niterations=100` explicitly to suppress it (or set it to any other value) is a fine escape hatch. Note this in the docstring.

(Alternative considered: keep `niterations: int = 100` and rely on `sklearn`'s `get_params` machinery to detect overrides. Too brittle; sklearn's introspection treats the default as a user value once `__init__` runs.)

### Emit the warning from `_validate_and_modify_params`

Inside `_validate_and_modify_params` (currently lines 1613-1701 in `pysr/sr.py`), right after the existing `maxsize` warning block (line 1648-1651), add:

```python
if self._niterations_was_default:
    warnings.warn(
        f"`niterations` is using the default of {_DEFAULT_NITERATIONS}, which "
        "is intentionally small for quick demos. For production runs you will "
        "almost always want a higher value (commonly 1000+). Set `niterations` "
        "explicitly to silence this warning. "
        "See https://ai.damtp.cam.ac.uk/pysr/tuning/ for guidance."
    )
```

`_validate_and_modify_params` is called from `fit` (`PySRRegressor.fit` at `pysr/sr.py` line 2411), so the warning fires on the actual search, not at `__init__`. This matches the existing `maxsize > 40` warning placement and avoids spamming users who construct a regressor without ever calling `.fit()` (e.g. doc generation).

### Docstring touch-up

Update the `niterations` docstring (around `pysr/sr.py` line 411-414) with one sentence:

> By default this is `100`, which is intentionally small for quick demos; production runs typically want a much larger value. PySR emits a warning if you leave this at the default; set it explicitly (even to `100`) to silence the warning.

### Test

Add to `pysr/test/test_main.py` (mirrors the existing `pytest.warns` usage pattern in that file):

```python
def test_warns_on_default_niterations():
    import warnings
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        model = PySRRegressor(
            niterations=100,  # explicit, same value as default -> still triggers
            # (documented edge case)
        )
        # Use a tiny dataset so fit returns fast; or stub _run.
        ...
    # Easier path: don't call .fit, call _validate_and_modify_params directly.
    model_default = PySRRegressor()
    with pytest.warns(UserWarning, match="niterations.*default"):
        model_default._validate_and_modify_params()

    model_explicit = PySRRegressor(niterations=500)
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        model_explicit._validate_and_modify_params()
    assert not any("niterations" in str(w.message) for w in caught)
```

Prefer the second form (direct `_validate_and_modify_params` call, no `.fit`) - the existing test file uses this pattern for the `maxsize` warning test, so it slots in cleanly and runs in <1s.

### PR body

> Towards #603.
>
> Adds a one-shot warning when `PySRRegressor` is run with the default `niterations=100`, pointing users at the tuning page. Setting `niterations` to any explicit value silences it.
>
> Detection uses an `__init__`-time equality check (`niterations == 100`) recorded as `_niterations_was_default`. One acknowledged edge: a user who explicitly passes `niterations=100` still triggers the warning; mentioned in the docstring with the escape hatch (set it to any other value, or accept the warning).
>
> Implementation mirrors the existing `maxsize > 40` warning at the top of `_validate_and_modify_params`, with a corresponding test in `pysr/test/test_main.py` that runs `_validate_and_modify_params` directly (no full search needed).

## Why this matters

`MilesCranmer/PySR#603` (authored by Miles 2024-04-19):

> I see a lot of people using the default niterations (purposefully small) for "production runs" and not getting as good results. Either the default `niterations` should be higher, or, unless set explicitly, should raise a warning and point the user to the tuning page.

The current default is `niterations=100` (`pysr/sr.py` line 956). The intent of the small default is to keep the README/getting-started example fast, but users routinely leave it at 100 for real searches and then report poor equation quality, which is on the maintainer's mind (see also #1001 "automated tips" and #1018 "smart default for parallelism").

This plan implements the warning option, not the breaking default change. The warning fires only when the user did NOT explicitly pass `niterations`, so deliberate users get no noise.

## Testing

- `pytest pysr/test/test_main.py::test_warns_on_default_niterations -q` passes.
- `pytest pysr/test/test_main.py -q` passes (no other warning-counter test gets a new spurious warning).
- `python -c "from pysr import PySRRegressor; PySRRegressor()._validate_and_modify_params()"` prints a single `UserWarning` mentioning `niterations` and the tuning URL.
- `python -c "from pysr import PySRRegressor; PySRRegressor(niterations=500)._validate_and_modify_params()"` prints no `niterations` warning.
- `python -c "from pysr import PySRRegressor; PySRRegressor(niterations=100)._validate_and_modify_params()"` prints the warning (documented edge case).
- `grep -n "_niterations_was_default\|_DEFAULT_NITERATIONS" pysr/sr.py` returns the new assignments and references; no other code paths read these names.
- `python -m mypy pysr/sr.py` reports no new errors (the sentinel constant is `int`, no annotation gymnastics).
- `ruff check pysr/sr.py pysr/test/test_main.py` reports no new findings.

Fixes #603

AI was used for assistance.

### DIFF
--- a/pysr/sr.py
+++ b/pysr/sr.py
@@ -342,6 +342,7 @@ def _validate_export_mappings(extra_jax_mappings, extra_torch_mappings):
 
 # Class validation constants
 VALID_OPTIMIZER_ALGORITHMS = ["BFGS", "NelderMead"]
+_DEFAULT_NITERATIONS = 100
 
 
 @dataclass
@@ -412,7 +413,10 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         Number of iterations of the algorithm to run. The best
         equations are printed and migrate between populations at the
         end of each iteration.
-        Default is `100`.
+        Default is `100`, which is intentionally small for quick demos;
+        production runs typically want a much larger value. PySR emits a
+        warning if you leave this at the default; set it to a different
+        value to silence the warning.
     populations : int
         Number of populations running.
         Default is `31`.
@@ -953,7 +957,7 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         unary_operators: list[str] | None = None,
         operators: dict[int, list[str]] | None = None,
         expression_spec: AbstractExpressionSpec | None = None,
-        niterations: int = 100,
+        niterations: int = _DEFAULT_NITERATIONS,
         populations: int = 31,
         population_size: int = 27,
         max_evals: int | None = None,
@@ -1068,6 +1072,7 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
         self.operators = operators
         self.expression_spec = expression_spec
         self.niterations = niterations
+        self._niterations_was_default = niterations == _DEFAULT_NITERATIONS
         self.populations = populations
         self.population_size = population_size
         self.ncycles_per_iteration = ncycles_per_iteration
@@ -1651,6 +1656,16 @@ class PySRRegressor(MultiOutputMixin, RegressorMixin, BaseEstimator):
             )
         elif self.maxsize < 7:
             raise ValueError("PySR requires a maxsize of at least 7")
+
+        if self._niterations_was_default:
+            warnings.warn(
+                f"`niterations` is using the default of {_DEFAULT_NITERATIONS}, "
+                "which is intentionally small for quick demos. For production "
+                "runs you will almost always want a higher value (commonly "
+                "1000+). Set `niterations` to a different value to silence this "
+                "warning. "
+                "See https://ai.damtp.cam.ac.uk/pysr/tuning/ for guidance."
+            )
 
         # NotImplementedError - Values that could be supported at a later time
         if self.optimizer_algorithm not in VALID_OPTIMIZER_ALGORITHMS:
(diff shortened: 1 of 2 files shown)


MilesCranmer:
Thanks! However, see https://github.com/MilesCranmer/PySR/issues/603#issuecomment-4493178186

PR #1210, opened 2026-06-22
fix: clarify TemplateExpressionSpec reload errors

### LINKED ISSUE #846: Not possible to load a model fitted with TemplateExpressionSpec
### Discussed in https://github.com/MilesCranmer/PySR/discussions/844

<div type='discussions-op-text'>

<sup>Originally posted by **ibengtsson** February 28, 2025</sup>
Hi!

First of all, thank you for a great library. 

I'm having some issues when working with TemplateExpresssionSpec, and it could very well be a fault on my end... If I create a model with a TemplateExpression, fit and save it, and then try and load it (with `from_file`) in a different script, I cannot do any model predictions. I just get the error:

```
ValueError: Failed to evaluate the expression. If you are using a custom operator, make sure to define it in `extra_sympy_mappings`, e.g., `model.set_params(extra_sympy_mappings={'inv': lambda x: 1/x})`, where `lambda x: 1/x` is a valid SymPy function defining the operator. You can then run `model.refresh()` to re-load the expressions.
```

I don't use any custom operators in my TemplateEpxressionSpec, but maybe I still need to add some mappings to make the loading work?

Grateful for any help!</div>

MilesCranmerBot:
Fixes #846.

- Detect pickled/reloaded `TemplateExpressionSpec` models
- Raise a targeted unsupported-reload error for predict/export paths
- Add regression coverage for the error message


MilesCranmer:
@MilesCranmerBot why doesn't this show up in coverage?

MilesCranmer:
Closing: the premise does not hold on current master. Fitted TemplateExpressionSpec models unpickle correctly as of 2.0.0-beta.3, verified by the fresh-process round-trip regression test (ca310584): fit, pickle, delete the run directory, then load and predict in a fresh process with prediction parity and lambda_format rebuilt from julia_state_stream_. The confusing errors were v1-only (1.5.x fails at dump time), long fixed upstream. Merging this PR would regress working reload behavior into a NotImplementedError. Tracking in #846.

PR #1280, opened 2026-08-07
feat: add runtime TypeSpec support

MilesCranmerBot:
## Summary

- add `TypeSpec(name, fields, sample, ...)` for symbolic regression over custom Julia value types
- compile each TypeSpec into one deterministic private Julia module per process, with exact-source reuse, replacement warnings, and conflicting checkpoint-restore protection
- make fitting transactional so failed searches and rejected warm starts leave the previously fitted Julia state intact
- preserve and replay operators, objectives, complexity mappings, early-stop conditions, templates, and worker definitions across serialization and multiprocessing
- support discrete constants through `mutate` and continuous constant optimization through `scalar_constants` and `with_scalar_constants`
- support typed `OperatorEnum` evaluation, full objectives, `ExpressionSpec`, explicit-keyword `TemplateExpressionSpec`, and custom expression specs that declare TypeSpec support
- remove the legacy positional `TemplateExpressionSpec` constructor and its obsolete compatibility machinery
- pin SymbolicRegression.jl `v2.0.0-beta.5` for custom-value mutation support

## Example

```python
type_spec = TypeSpec(
    name="Vec2",
    fields={"data": "Vector{Float64}"},
    sample="rng -> Vec2(randn(rng, 2))",
    scalar_constants="value -> value.data",
    with_scalar_constants="(value, constants) -> Vec2(constants)",
)

model = PySRRegressor(
    type_spec=type_spec,
    operators={
        1: ["rotate90(x::Vec2) = Vec2([-x.data[2], x.data[1]])"],
        2: ["add_vectors(x::Vec2, y::Vec2) = Vec2(x.data + y.data)"],
    },
    elementwise_loss=(
        "vector_loss(x::Vec2, y::Vec2)::Float64 = "
        "sum(abs2, x.data - y.data)"
    ),
)
```

The documentation includes custom full objectives plus vector-valued, string-valued, heterogeneous tensor, variable-length constant, template, multiprocessing, and checkpoint examples.

## Verification

- `44 passed, 36 subtests passed` in `pysr/test/test_type_specs.py`
- `157 passed, 38 subtests passed` in `pysr/test/test_main.py`
- changed-file pre-commit hooks passed
- `git diff --check` passed


pysr/sr.py
```diff
@@ -2342,6 +2453,11 @@ def _run(
             if self.worker_imports is not None
             else None
         )
+        loss_type_kw = (
+            {"loss_type": jl.seval(self.loss_type)}
+            if self.loss_type is not None
+            else {}
+        )
```
MilesCranmer: jl.seval(None) is fine, you don't need all this

pysr/type_specs.py
```diff
@@ -0,0 +1,140 @@
...
+    fields: dict[str, str] | None = None
+    init_value: str | None = None
+    sample_value: str | None = None
+    mutate_value: str | None = None
+    count_scalar_constants: int | str | None = None
+    can_optimize: bool | None = None
+
+    _installed: ClassVar[dict[str, tuple[tuple[Any, ...], AnyValue]]] = {}
+
+    def install(self) -> AnyValue:
```
MilesCranmer: install feels wrong... idk. what about instantiate?

MilesCranmer: Or... `eval`?

pysr/type_specs.py
```diff
@@ -0,0 +1,140 @@
...
+        expected = {"init": (0,), "sample": (1, 2), "mutate": (3, 4), "count": (1,)}[
+            kind
+        ]
+        if arity not in expected:
+            raise ValueError(
+                f"{kind}_value must accept {expected}; got {arity} arguments."
+            )
+
+        installers = {
+            "init": "(T, f) -> @eval SymbolicRegression.init_value(::Type{$T}) = $f()",
```
MilesCranmer: Why are these inside `@eval`? Why not just `seval` directly?

pysr/sr.py
```diff
@@ -2042,6 +2131,10 @@ def _run(
         parallelism, numprocs = _map_parallelism_params(
             self.parallelism, self.procs, getattr(self, "multithreading", None)
         )
+        if self.type_spec is not None and parallelism != "serial":
+            raise NotImplementedError(
```
MilesCranmer: This should not be the case... At the very least, multithreading should work.

MilesCranmer (changes requested):

pysr/sr.py
MilesCranmerBot: Simplified in 9601b5a6: the keyword handling is now a single inline conditional, with no temporary dict. I kept the keyword omitted when absent because the pinned backend rejects an explicit Julia `nothing` for `loss_type`; the normal numeric regression test covers that path.

pysr/type_specs.py
MilesCranmerBot: Renamed to `instantiate` in 9601b5a6.

pysr/type_specs.py
MilesCranmerBot: Replaced the `@eval` wrappers with direct `jl.seval` method definitions in 9601b5a6.

pysr/sr.py
MilesCranmerBot: Removed the serial-only restriction and added a multithreaded TypeSpec fit/predict regression test in 9601b5a6.

docs/src/examples.md
```diff
@@ -318,118 +318,79 @@ to make predictions with the most accurate expression.
...
-then `T` will be `ComplexF32` and `L` will be `Float32` (since we need to return
-a real number from the loss function). But, you don't need to worry about this, just
-make sure to return a scalar number of type `L`.
-
-The `tree` argument is the current expression being evaluated. You can read
-about the `tree` fields [here](https://ai.damtp.cam.ac.uk/symbolicregression/stable/types/).
+With automatic batching, a three-argument objective receives a dataset already
+restricted to the active batch. A four-argument objective receives the full
+dataset and the selected row indices. This makes sample-aligned auxiliary data
+available without copying it into every batch.
```
MilesCranmer: Delete "This makes sample-aligned auxiliary data available without copying it into every batch." - not even true. Instead we can write "This can be useful when needing to do custom batching operations."

MilesCranmer (dismissed):

MilesCranmer (approved):

PR #1282, opened 2026-08-09
feat: expose new mutation / plugin interface + update to v2.0.0-beta.3

MilesCranmerBot:
Updates SymbolicRegression.jl to v2.0.0-beta.1 and exposes its mutation and plugin configuration through PySR.

- Mirror the backend `default_mutations` and `mutations` interfaces.
- Mirror the backend `default_plugins` and `plugins` interfaces.
- Preserve existing `weight_*` behavior and add `weight_backsolve`.
- Update the crossover default to `0.2`.

Tests: focused mutation, plugin, compatibility, and parameter-grouping tests; pre-commit; mypy.


docs/src/options.md
MilesCranmer, line 1: Not sure we need to describe this as an option. It's a little weird. People can find it through the API page I guess? The options is for common use-cases... Not for low level plugin configurations.

pysr/mutations.py
```diff
@@ -0,0 +1,113 @@
...
+
+@dataclass(frozen=True)
+class ConstantMutation(AbstractMutation):
+    """Perturb a constant.
+
+    Settings left as ``None`` use the SymbolicRegression.jl defaults.
+    """
+
+    perturbation_factor: float | None = None
+    probability_negate: float | None = None
```
MilesCranmer: Why not just set it here, so that users can see the values?

pysr/sr.py
```diff
@@ -1464,6 +1512,15 @@ def __getstate__(self) -> dict[str, Any]:
                 ]
         return pickled_state
 
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        schema_version = state.pop("_checkpoint_schema_version", None)
+        if schema_version != _CHECKPOINT_SCHEMA_VERSION:
```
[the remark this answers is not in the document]
MilesCranmer: (Choosing to ignore)

MilesCranmer (approved):

PR #1371, opened 2026-09-16
docs: keep <details> headings on the summary line

MilesCranmer:
The Docker, Apptainer and Troubleshooting install sections on the docs homepage showed their title on its own line under the disclosure marker. VitePress renders the heading inside `<summary>` as a block element.

CSS-only fix: render a summary heading inline. Markdown headings are left alone, so the page outline, heading slugger dedup (`docker` / `docker-1`) and permalink anchors keep working. Scoped with `summary:has(> h3)` so plain `<details>` blocks in the example pages are untouched.

Verified on a `vitepress build`: three closed `<details>` elements with `<h3 id="docker">` intact plus its `header-anchor`, and summary and heading on the same line.


MilesCranmer:
All three Codex comments applied: the generator change is dropped entirely, so headings stay real Markdown headings (outline, anchors and slug dedup preserved), and the spacing rules are scoped with `summary:has(> h3)` instead of every `.vp-doc details`. Now CSS only.

PR #1379, opened 2026-09-20
feat: enable concurrent GC sweeping by default

MilesCranmerBot:
Sets `JULIA_NUM_GC_THREADS` to half the CPU count plus one concurrent sweep thread, alongside the other juliacall defaults in `julia_import.py`. A user's own `JULIA_NUM_GC_THREADS` is left untouched.

The search is GC-bound at high thread counts. Measured on SymbolicRegression.jl's default preset (96 populations x 27, 100 iterations, plain `Node`, Julia 1.12, Genoa node), env-var form vs Julia's default:

| threads | default | with concurrent sweep |
|---|---|---|
| 8 | 31.2 ± 0.4 s | 31.5 ± 0.4 s |
| 32 | 11.8 ± 0.4 s | 10.8 ± 0.0 s |
| 96 | 8.4 s | 6.3 s |

Same Pareto front and best loss in every arm. The `N,1` form is accepted by Julia 1.10, 1.11, and 1.12. Verified through juliacall on an 8-core machine: `Threads.ngcthreads() == 5`; with `JULIA_NUM_GC_THREADS=2,0` set beforehand Julia sees 2.


MilesCranmer:
@codex review

